PR #13, opened 2023-11-23
fix the issue of model repetition

t0ng7u:
fix the issue of model repetition


xiaotian044500:
本地 Agent 已接管，开始处理 #13...

PR #259, opened 2024-05-16
Add Baidu Default Behavior and Updating Baidu&360 Models & Prices

jimmyshjj:
Improving the default behavior of Baidu, adding new Baidu and 360 models and prices based on official documentation.


Calcium-Ion:
close #246

PR #710, opened 2025-01-21
Fix temperature not being set to 0 due to json omitempty

hubutui:
The issue was caused by the `omitempty` tag in the Go struct, which prevented the `temperature` field from being included in the JSON output when it was set to 0.


Calcium-Ion:
不能这样修改，有的模型不支持0 temperature，只能对单个模型进行调整

hubutui:
how about this? I make `Temperature` `*float64`, if it's `nil`, it will be omitted too.

PR #723, opened 2025-02-04
Support for MokaAI M3E

kuwork:
Support for MokaAI M3E and fixed problem : chanel test did not refresh


Calcium-Ion:
感谢大佬pr，这是大重构了啊

PR #732, opened 2025-02-07
feat: Add FIM support with /completions for relay siliconflow

jyc001:
Add FIM support with /completions for relay siliconflow
Add case RelayModeCompletions in response handler

如题
为siliconflow渠道添加/completions路由支持 为支持其FIM功能


ghost:
改错了 等下重新提交

PR #903, opened 2025-03-26
Access Tongyi Wanxiang text-to-video and image-to-video (currently not implemented for billing).

asjfoajs:
This is divided into three parts to implement:
1. Asynchronously call the video generation interface and insert the task ID into the task table.
2. Use a scheduled task to periodically call the interface to retrieve the results and update the task table.
3. Implement an HTTP interface to query the task table for the status and results.


asjfoajs:
关闭

PR #1136, opened 2025-05-31
feat: support tolerant unmarshal in common.DecodeJson

jtsang4:
### Problem
- Issue detail: https://github.com/QuantumNous/new-api/issues/1131
- API responses sometimes return `created` field as float64 (e.g., `1748724739.7194004`) 
- Go struct expects `int64` type, causing unmarshaling error: `cannot unmarshal number 1748724739.7194004 into Go struct field OpenAITextResponse.created of type int64`

### Solution
- Implemented custom `int64Decoder` using existing `jsoniter` library
- Automatically converts float64 to int64 during JSON unmarshaling
- Global effect - applies to all `int64` fields across the codebase

### Features
- ✅ Handles float64 → int64 conversion (e.g., `1748724739.7194004` → `1748724739`)
- ✅ Preserves existing integer values unchanged
- ✅ Supports string numbers as fallback (e.g., `"1748724739.123"` → `1748724739`)
- ✅ Maintains backward compatibility
- ✅ Content strings containing "created" are unaffected

### Changes
- Modified `common/json.go` to use custom jsoniter decoder
- Leverages existing `github.com/json-iterator/go` dependency

### Testing
Verified with various test cases including real API responses, edge cases, and content validation.

### DIFF
--- a/common/json.go
+++ b/common/json.go
@@ -2,17 +2,60 @@ package common
 
 import (
 	"bytes"
-	"encoding/json"
+	"unsafe"
+
+	jsoniter "github.com/json-iterator/go"
 )
 
+var (
+	// Create a custom jsoniter instance with fault-tolerant configuration
+	tolerantJson = jsoniter.Config{
+		EscapeHTML:             false,
+		SortMapKeys:            false,
+		ValidateJsonRawMessage: true,
+		UseNumber:              false,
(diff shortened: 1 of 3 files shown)


Calcium-Ion:
建议单独适配特殊渠道，而不是整体修改DecodeJson函数

PR #1824, opened 2025-09-16
Stripe checkout add customer full address details during checkout

sanjibnarzary:
Some country like India require customers' full details during. With this PR its possible during checkout with stripe.

## Summary by CodeRabbit

* **New Features**
  * Added Creem as a new top-up payment method with preset products, checkout flow, webhook handling, and automated crediting after successful payment.
  * Added Creem settings UI (API Key, Webhook Secret, Test Mode) and product manager in Payment Settings.
  * TopUp page now shows Creem products, confirmation modal, and initiates Creem checkout.

* **Improvements**
  * Status API exposes Creem availability and product list for dynamic UI.
  * Stripe checkout: English-only messages, collects shipping, required billing address, phone, and updated success/cancel URLs.

### DIFF
--- a/controller/topup_stripe.go
+++ b/controller/topup_stripe.go
@@ -34,37 +34,37 @@ type StripeAdaptor struct {
 }
 
 func (*StripeAdaptor) RequestAmount(c *gin.Context, req *StripePayRequest) {
-	if req.Amount < getStripeMinTopup() {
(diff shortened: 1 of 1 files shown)


wzxjohn:
I don't think forcing the collection of user information is a good idea.

PR #2002, opened 2025-10-10
fix(channel): handle dynamic frequency updates

qixing-jk:
- replace infinite sleep loop with time.Ticker to avoid goroutine leaks
- add immediate initial test execution before ticker starts
- implement frequency change detection and ticker recreation
- ensure proper ticker cleanup when loop exits or feature disabled

## Summary by CodeRabbit

* **Bug Fixes**
  * Frequency is re-read each inner loop iteration so scheduling adapts to changes immediately.
  * Interval values are logged for each run and a completion message is logged after each automated channel test.
  * Loop now checks at runtime and stops promptly if automatic channel testing is disabled, preventing blocking.

### DIFF
--- a/controller/channel-test.go
+++ b/controller/channel-test.go
@@ -624,8 +624,8 @@ func AutomaticallyTestChannels() {
 			}
 			for {
 				frequency := operation_setting.GetMonitorSetting().AutoTestChannelMinutes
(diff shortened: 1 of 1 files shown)


seefs001:
感觉你搞复杂了，直接把那个获取时间的和日志挪进去内层循环不就可以嘛

PR #2133, opened 2025-10-30
fix: root page does not have analytic code

wzxjohn:
The root page `https://xxx.xxx/` is served directly from the embed file system, which always delivers the original version of the index file — one that hasn’t been patched with the analytics code.

This PR prevents the static file system from serving the root page directly, ensuring that the version with the analytics code is always used.

## Summary by CodeRabbit

* **Refactor**
  * Improved embedded file system implementation for better maintainability and flexibility.
* **Bug Fixes**
  * Fixed handling of the root path when serving embedded files so requests to "/" behave correctly and return appropriate errors.


wzxjohn:
@coderabbitai generate docstrings

PR #2213, opened 2025-11-11
feat: support anthropic beta for AWS and Vertex

prnake:
Vertex: 透传 anthropic-beta
AWS: 从 header 中取出 anthropic-beta，设置到 body.anthropic_beta

## Summary by CodeRabbit

* **Enhancements**
  * The anthropic-beta header is now propagated into AWS Bedrock requests; when supported, requests are augmented so beta flags are forwarded to Bedrock.
  * The anthropic-beta header is also forwarded through Vertex AI requests, enabling consistent access to beta features across integrations.


prnake:
fixed in https://github.com/QuantumNous/new-api/commit/e1a52f1d5a4d95d91a1beb461c2e03ccadde93bc

PR #2501, opened 2025-12-23
fix(channel-test): responses fallback + better upstream error messages

FlowerRealm:
Problem:
- Some upstreams are responses-only (reject chat.completions 'messages'), so channel test fails even though the channel works via /v1/responses.
- Some upstream error payloads have no error.message (or only a 'detail' field), leading to empty UI messages like '模型 xxx:'.

Fix:
- Parse GeneralErrorResponse.detail and fallback to raw error JSON when message is missing.
- Ensure RelayErrorHandler returns a non-empty error even on nil/failed body reads.
- In auto-detect mode, retry via Responses API when chat.completions reports 'Unsupported parameter: messages'.
- Make Responses test request use stream=true with valid input.

Tests:
- Add unit tests for error parsing and RelayErrorHandler / NewAPIError.Error() fallbacks.

## Summary by CodeRabbit

* **New Features**
  * Default routing and playground/tools now use the Responses (/v1/responses) API path and payload format.
  * UI and endpoint options updated to reflect the Responses endpoint.

* **Bug Fixes**
  * Improved error-message extraction with a new detail fallback.
  * Safer handling when upstream response bodies are nil or unreadable.
  * Validation relaxed to allow Responses requests without an Input field.

* **Tests**
  * Added tests for error parsing, nil/unreadable bodies, and request validation.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/relay/channel/claude/relay-claude.go
+++ b/relay/channel/claude/relay-claude.go
@@ -483,9 +483,11 @@ func StreamResponseClaude2OpenAI(reqMode int, claudeResponse *dto.ClaudeResponse
 				}
 			}
 		} else if claudeResponse.Type == "message_delta" {
-			finishReason := stopReasonClaude2OpenAI(*claudeResponse.Delta.StopReason)
-			if finishReason != "null" {
-				choice.FinishReason = &finishReason
+			if claudeResponse.Delta != nil && claudeResponse.Delta.StopReason != nil {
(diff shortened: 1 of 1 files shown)


FlowerRealm:
@coderabbitai generate docstrings

PR #2511, opened 2025-12-24
Solve Claude Code crash issue

### LINKED ISSUE #2499: Claude Code crash when processing the response
**Routine Checks**

[//]: # (Remove the space in the box and fill with an x)
+ [ ] I have confirmed there are no similar issues currently
+ [ ] I have confirmed I have upgraded to the latest version
+ [ ] I have thoroughly read the project README, especially the FAQ section
+ [ ] I understand and am willing to follow up on this issue, assist with testing and provide feedback 
+ [ ] I understand and acknowledge the above, and understand that project maintainers have limited time and energy, **issues that do not follow the rules may be ignored or closed directly**

**Issue Description**
I have setup up a channel with local hosted openai service (SGLang + Deepseek), and It seems like claude code work for normal chat, but not tool use such as create/edit/list file. But when I switch it to another litellm proxy, claude code works well. So I think there must be something wrong with the response of new-api.

**Steps to Reproduce**
1. Setup a channel with local hosted openai-compatibile LLM service (such as vLLM or SGLang), select `openai` as the channel type, and configure base_url, fetch models.
2. Configure Claude Code to use new-api (export ANTHROPIC_BASE_URL, export ANTHROPIC_AUTH_TOKEN, ANTHROPIC_MODEL)
3. input something like "Create a Python hello world program", and wait

**Expected Result**
Claude code works well, and no crashes anymore.

**Related Screenshots**

The following is the console output of claude code
```text
I'll create a simple "Hello World" Python file for you.

✽ Generating… (esc to interrupt · 2s · ↓ 105 tokens · thinking)

──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
>  
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
  -- INSERT --                                                                                                           3858 tokens
This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason:
TypeError: Cannot read properties of undefined (reading 'startsWith')
    at tg5 (file:///home/lk/.nvm/versions/node/v22.5.1/lib/node_modules/@anthropic-ai/claude-code/cli.js:2207:11991)
    at CX1 (file:///home/lk/.nvm/versions/node/v22.5.1/lib/node_modules/@anthropic-ai/claude-code/cli.js:2207:12243)
    at CX1.next (<anonymous>)

──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
>  

```
If none, please delete this section.

JerryKwan:
fix #2499 
add stricter validation for content_block_start corresponding to tool call
and fix the crash issue when Claude Code is processing tool call

## Bug Fixes
- Fixed tool call processing to prevent emission of incomplete tool blocks when function names or arguments are missing, improving reliability and reducing unnecessary processing.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/service/convert.go
+++ b/service/convert.go
@@ -389,25 +389,29 @@ func StreamResponseOpenAI2Claude(openAIResponse *dto.ChatCompletionsStreamRespon
 				}
 
 				idx := blockIndex
-				claudeResponses = append(claudeResponses, &dto.ClaudeResponse{
-					Index: &idx,
-					Type:  "content_block_start",
-					ContentBlock: &dto.ClaudeMediaMessage{
-						Id:    toolCall.ID,
-						Type:  "tool_use",
-						Name:  toolCall.Function.Name,
-						Input: map[string]interface{}{},
-					},
-				})
-
-				claudeResponses = append(claudeResponses, &dto.ClaudeResponse{
-					Index: &idx,
-					Type:  "content_block_delta",
-					Delta: &dto.ClaudeMediaMessage{
-						Type:        "input_json_delta",
-						PartialJson: &toolCall.Function.Arguments,
-					},
-				})
+				if toolCall.Function.Name != "" {
+					claudeResponses = append(claudeResponses, &dto.ClaudeResponse{
+						Index: &idx,
+						Type:  "content_block_start",
+						ContentBlock: &dto.ClaudeMediaMessage{
+							Id:    toolCall.ID,
+							Type:  "tool_use",
+							Name:  toolCall.Function.Name,
+							Input: map[string]interface{}{},
+						},
+					})
+				}
+				
+				if len(toolCall.Function.Arguments) > 0 {
+					claudeResponses = append(claudeResponses, &dto.ClaudeResponse{
+						Index: &idx,
+						Type:  "content_block_delta",
+						Delta: &dto.ClaudeMediaMessage{
+							Type:        "input_json_delta",
+							PartialJson: &toolCall.Function.Arguments,
+						},
+					})
+				}
 
 				info.ClaudeConvertInfo.Index = blockIndex
 			}


service/convert.go
```diff
@@ -389,25 +389,29 @@ func StreamResponseOpenAI2Claude(openAIResponse *dto.ChatCompletionsStreamRespon
...
+						ContentBlock: &dto.ClaudeMediaMessage{
+							Id:    toolCall.ID,
+							Type:  "tool_use",
+							Name:  toolCall.Function.Name,
+							Input: map[string]interface{}{},
+						},
+					})
+				}
+				
+				if len(toolCall.Function.Arguments) > 0 {
```
seefs001, line 405: I'm guessing if there might be tools that don't have Arguments parameters but are still considered tools.

JerryKwan: I think tools without parameters do exist such as "get_current_time". Since there are no parameters, there should be no need to send an additional empty input_json_delta event.

seefs001:
Is this modification working well in your scenario? I can use GPT normally with both the modified version and the pre-modified version, so I can't verify it.

JerryKwan:
> Is this modification working well in your scenario? I can use GPT normally with both the modified version and the pre-modified version, so I can't verify it.

Yes, It works in my scenario.
I have done some tests using SGLang + Deepseek / Kimi-K2-Thinking, all of them work as expected. But the pre-modified version caused Claude Code crash.

seefs001:
> > Is this modification working well in your scenario? I can use GPT normally with both the modified version and the pre-modified version, so I can't verify it.
> 
> Yes, It works in my scenario. I have done some tests using SGLang + Deepseek / Kimi-K2-Thinking, all of them work as expected. But the pre-modified version caused Claude Code crash.

Thank you for your thorough research. I will merge it, and it is expected to be included in the next version.

PR #2706, opened 2026-01-21
fix(claude): 检测并报错流中断导致的空响应

XiaoLFeng:
添加 `claudeResponseText` 函数以统一提取 Claude 响应文本。
在流式和非流式处理结束前校验响应内容，若响应为空且使用率为零，
则返回 BadResponseBody 错误，避免流中断导致的静默失败。

## Summary by CodeRabbit

* **New Features**
  * Added a global setting and UI controls to enable/disable empty-stream retry and configure retry delay (enabled by default, default delay 100ms).
* **Bug Fixes**
  * Better detection of empty/invalid streaming responses; will surface a clear error and trigger retry behavior when configured.
* **Localization**
  * Added translations for the new "Empty-stream retry (empty output and token=0)" setting and its delay across supported locales.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #2737, opened 2026-01-24
feat: automatic retry and disable for empty replies

seefs001:
-> #2706

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added two new configuration toggles to handle empty API responses: automatic retry on empty response and automatic channel disable on empty response
  * Added UI switches in settings to enable or disable empty response handling behavior

* **Bug Fixes**
  * Improved handling of malformed and empty responses from API providers
  * Enhanced error logging for JSON parsing failures with better context and truncation
  * Fixed downstream connection state tracking to prevent erroneous data transmission during streaming

* **Improvements**
  * Better context cancellation handling across streaming operations


WhiteRaven11:
It would also be great if repeated requests could be configured according to the status on the log, that is, if there is an "Error". Because Nano Banana sometimes makes a text refusal for images, but does not give images, and when you request the same request again, it may already give you an image.

seefs001:
> It would also be great if repeated requests could be configured according to the status on the log, that is, if there is an "Error". Because Nano Banana sometimes makes a text refusal for images, but does not give images, and when you request the same request again, it may already give you an image.

现在有可以配置自动重试的状态码，如果你能确定你的上游的这种情况下返回的状态码直接配置就可以吧，和这个PR做的内容无关

xiaolfeng:
目前没有遇到什么问题（需求够的）

seefs001:
> 目前没有遇到什么问题（需求够的）

咦你在用我的这个PR吗？

xiaolfeng:
> > 目前没有遇到什么问题（需求够的）
> 
> 咦你在用我的这个PR吗？

在 #2706 这里你说 link 一下看看的

WenHao0804:
这个PR太有用了，刚发现流中断导致无响应的bug，准备提个issue来着

WhiteRaven11:
Please add the following features:

1. Sometimes gemini\vertex is sent text instead of an image, and if the price of the model is set for a call and not for tokens, the user's funds are debited. However, it often happens that a user sends a request to take a photo of himself and Google rejects it as a violation of the rules. 

Please add an option so that you can set up the channel for the image so that if Google sends text instead of an image, the site does not charge funds from it. 

With these settings, in order to 

2. If Google returns text instead of an image, so that we can resend the request. 

(It often happens when resending, successfully)

WhiteRaven11:
3. Another important edit. When a user sends a request, the site shows errors in the log, even if a system of repeated requests was used, where the final result is successful. 



You can make sure that the erroneous requests are seen by admins only and not by users. It is only about those who, upon repeated request, become successful in the end. 


And for the admins, how to indicate that this error was worked out upon repeated request?

WhiteRaven11:
@seefs001

KingingWang:
这个pr还需要测试什么吗?什么时候可以合并呢?

PR #2758, opened 2026-01-27
feat: async log buffer to prevent DB blocking during error floods

KiGamji:
## Why

We had a production incident where a short upstream outage (Vertex/Gemini `unexpected EOF`) caused a burst of failures.
Each failure triggered a synchronous INSERT into the `logs` table. Under burst load this created a thundering herd of
DB writes, saturated the DB connection pool, and cascaded into service unresponsiveness / restart.

This PR prevents error-flood logging from becoming a second outage.

## What

- Switch log writes for `RecordErrorLog` / `RecordConsumeLog` to an async, in-memory buffer.
- A background worker flushes buffered log rows to the DB on a fixed interval (default: 1s).
- This decouples request handling from DB latency, so upstream error bursts don’t block handlers or exhaust DB connections.

## Impact

- Much lower DB pressure during bursts: handlers enqueue quickly instead of blocking on INSERTs.
- Service stays responsive even when upstream providers flap.
- Logs may appear in the UI with up to ~flush-interval delay (default ~1s).

## Configuration

- `ASYNC_LOG_ENABLED=false` — disable async buffering (fallback to sync behavior).
- `LOG_FLUSH_INTERVAL` — flush interval in seconds (default: 1).

## Notes / trade-offs

- Uses an unbounded in-memory queue: in extreme sustained error floods memory can grow.
  (Trade-off chosen to avoid reintroducing sync blocking)
- Flush on shutdown attempts to drain pending logs before DB close.

## Summary by CodeRabbit

* **New Features**
  * Asynchronous, in-memory log buffering for non-blocking logging with configurable flush intervals.
  * Graceful shutdown handling to ensure buffered logs are flushed on exit.

* **Documentation**
  * Added environment variable examples to configure async logging and flush intervals.


KiGamji:
@Calcium-Ion Hey there, could you review my PRs please? Appreciate it 🙏

KiGamji:
@Calcium-Ion

PR #2968, opened 2026-02-18
fix: normalize input_text content blocks in Claude-to-OpenAI conversion

0-don:
- Map `input_text` → `text` in `ClaudeToOpenAIRequest` so clients sending Responses API content types via `/v1/messages` don't get silently dropped
- Fixes `Invalid value: 'input_text'` errors from upstream providers

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of text messages so content is converted consistently across formats.
  * Message text now distinguishes sender role (assistant vs. others) to ensure correct interpretation of incoming and outgoing text.

### DIFF
--- a/service/convert.go
+++ b/service/convert.go
@@ -127,7 +127,7 @@ func ClaudeToOpenAIRequest(claudeRequest dto.ClaudeRequest, info *relaycommon.Re
 
 			for _, mediaMsg := range contents {
 				switch mediaMsg.Type {
-				case "text":
+				case "text", "input_text":
 					message := dto.MediaContent{
 						Type:         "text",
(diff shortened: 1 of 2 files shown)


seefs001:
应该是搞错地方了，claude文档上没有这一个类型，这个报错可能是别的地方的转换错误

0-don:
input_text is not in the Anthropic spec, it's from the Responses API. OpenClaw sends it through /v1/messages when using codex models. Known bugs on their side (openclaw/openclaw#13189, openclaw/openclaw#18787). Same structure as text, just a different type string. Without this, the content gets silently dropped and upstream returns errors.

PR #3105, opened 2026-03-03
feat: cross-domain auth support for multi-subdomain deployments

0-don:
## Summary

Adds optional cross-domain session cookie support so deployments with multiple subdomains (e.g. `api.example.com` and `example.com`) can share authentication state. All changes are opt-in and backwards compatible.

### Changes

- **Configurable cookie domain** (`SESSION_COOKIE_DOMAIN` env var): When set (e.g. `.example.com`), the session cookie is shared across all subdomains. When not set, behavior is identical to before (cookie scoped to exact host, SameSite Strict, Secure false)
- **CORS on API router**: Adds `middleware.CORS()` to the `/api` route group. The relay and dashboard routers already had it, but auth endpoints (`/api/user/login`, `/api/status`, `/api/oauth/*`) were missing CORS headers, blocking cross-origin requests with credentials
- **Password auth status in /api/status**: Adds `password_login_enabled` and `password_register_enabled` fields so frontends can dynamically show/hide the password form
- **OAuth redirect param**: `/api/oauth/state` accepts an optional `redirect` query param. After successful OAuth login, the backend redirects to this URL instead of returning JSON. The redirect URL is validated against the `ServerAddress` root domain. When no redirect param is provided, the existing JSON response is returned (no change for the built-in frontend)
- **Refactored `setupLogin`**: Split into `setupSession` (saves session only) + `setupLogin` (saves session and returns JSON). This allows the OAuth redirect flow to set the session without writing a JSON response

### Backwards Compatibility

Every change is opt-in or additive:
- No env var set = zero behavior change
- New `/api/status` fields are purely additive (existing consumers ignore unknown fields)
- OAuth redirect only activates when `redirect` param is explicitly passed
- CORS was already enabled on relay/dashboard routers with the same config

## Summary by CodeRabbit

* **New Features**
  * Added password login/register status indicators
  * OAuth cross-domain redirect support with stricter redirect validation
  * Dynamic session cookie domain configuration for flexible deployments

* **Improvements**
  * Consolidated session setup with better error handling during login
  * Added CORS middleware for API routes and preserved permissive origin behavior

### DIFF
--- a/middleware/cors.go
+++ b/middleware/cors.go
@@ -8,7 +8,7 @@ import (
 
 func CORS() gin.HandlerFunc {
 	config := cors.DefaultConfig()
-	config.AllowAllOrigins = true
+	config.AllowOriginFunc = func(origin string) bool { return true }
 	config.AllowCredentials = true
 	config.AllowMethods = []string{"GET", "POST", "PUT", "DELETE", "OPTIONS"}
 	config.AllowHeaders = []string{"*"}


0-don:
No longer needed. Using a Next.js proxy route instead, which avoids all cross-domain cookie and CORS issues without requiring backend changes.

PR #3237, opened 2026-03-12
feat(openai): add responses reasoning input compatibility flag

bkidy:
## Summary
- add a default-off `REMOVE_RESPONSES_REASONING_INPUT` flag for OpenAI Responses requests
- strip `reasoning*` items from `input` when the flag is enabled before forwarding upstream
- help partially compatible relays avoid account/session-bound errors such as `Item with id 'rs_...' not found`

## Why
Some OpenAI-compatible relays can serve `/v1/responses`, but do not fully support account/session-bound reasoning state across turns. In those cases, replaying reasoning items from prior responses may fail with errors like:
- `Item with id 'rs_...' not found`
- `invalid_encrypted_content`

This PR adds a small compatibility escape hatch without changing the default behavior.

## Behavior
- default remains unchanged
- when `REMOVE_RESPONSES_REASONING_INPUT=true`, `reasoning*` items are removed from the Responses `input` array before the request is sent upstream
- malformed input items are preserved as-is

## Trade-offs
This compatibility mode may reduce some Responses-native continuity/cost benefits because account-bound reasoning items are no longer replayed.

## Release Notes

* **New Features**
  * Added configurable option to remove reasoning data from responses via environment variable.
  * Integrated filtering mechanism with error handling for response processing.

### DIFF
--- a/common/init.go
+++ b/common/init.go
@@ -143,6 +143,8 @@ func initConstantEnv() {
 	constant.GenerateDefaultToken = GetEnvOrDefaultBool("GENERATE_DEFAULT_TOKEN", false)
 	// 是否启用错误日志
 	constant.ErrorLogEnabled = GetEnvOrDefaultBool("ERROR_LOG_ENABLED", false)
+	// 是否移除 Responses 请求中与账号/会话绑定的 reasoning 输入项
+	constant.RemoveResponsesReasoningInput = GetEnvOrDefaultBool("REMOVE_RESPONSES_REASONING_INPUT", false)
 	// 任务轮询时查询的最大数量
 	constant.TaskQueryLimit = GetEnvOrDefault("TASK_QUERY_LIMIT", 1000)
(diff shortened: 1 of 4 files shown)


bkidy:
I hit this in production with OpenAI Responses routed through a partially compatible relay.
The relay can serve `/v1/responses`, but replaying account/session-bound reasoning items across turns may fail with errors like:
- `Item with id 'rs_...' not found`
- `invalid_encrypted_content`
This PR keeps default behavior unchanged and adds a small compatibility escape hatch:
- disabled by default
- enabled only via `REMOVE_RESPONSES_REASONING_INPUT=true`
- strips `reasoning*` items from the Responses `input` array before forwarding upstream
I kept the change intentionally small to avoid affecting the normal OpenAI Responses path.

RedwindA:
建议研究并使用`渠道亲和性`功能

PR #4084, opened 2026-04-04
fix(token): add batch keys endpoint to prevent 429 rate limit on bulk copy

ImogeneOctaviap794:
## Problem

When users batch-copy 50+ token keys from the token management page, the frontend fires N individual `POST /api/token/:id/key` requests simultaneously via `Promise.all`. The `CriticalRateLimit` (20 req/20min per IP) rejects most of them with HTTP 429, **making the batch copy feature completely broken**.

## Solution

Add a single `POST /api/token/batch/keys` endpoint that accepts an array of token IDs and returns all keys in **one request**, eliminating the N-request problem entirely.

## Changes (5 files, +63 / -6)

| File | Change |
|------|--------|
| `model/token.go` | Add `GetTokensByIdsAndUserId` for batch DB query |
| `controller/token.go` | Add `BatchGetTokenKeys` handler (max 100 IDs) |
| `router/api-router.go` | Register `POST /api/token/batch/keys` route |
| `web/src/helpers/token.js` | Add `fetchTokenKeysBatch` calling batch endpoint |
| `web/src/hooks/tokens/useTokensData.jsx` | Rewrite `batchCopyTokens` to use single batch request |

## Security

- Retains `CriticalRateLimit` + `UserAuth` middleware on the new endpoint
- Only returns keys owned by the authenticated user (`WHERE user_id = ?`)
- Hard cap of 100 IDs per request

## Before vs After

| | Before | After |
|---|---|---|
| Copy 50 keys | 50x `POST /api/token/:id/key` | **1x** `POST /api/token/batch/keys` |
| Rate limit consumption | 50 (triggers 429) | **1** (no issue) |

## Summary by CodeRabbit

* **New Features**
  * Added batch token key retrieval API endpoint supporting up to 100 tokens per request
  * Optimized multi-token operations to use single API call instead of individual requests
  * Implemented input validation to prevent invalid batch requests

### DIFF
--- a/controller/token.go
+++ b/controller/token.go
@@ -94,6 +94,37 @@ func GetTokenKey(c *gin.Context) {
 	})
 }
 
+func BatchGetTokenKeys(c *gin.Context) {
+	userId := c.GetInt("id")
+	var req struct {
+		Ids []int `json:"ids"`
+	}
+	if err := common.DecodeJson(c.Request.Body, &req); err != nil {
+		common.ApiError(c, err)
+		return
+	}
+	if len(req.Ids) == 0 {
+		common.ApiError(c, fmt.Errorf("ids is empty"))
+		return
+	}
+	if len(req.Ids) > 100 {
+		common.ApiError(c, fmt.Errorf("batch size exceeds limit (max 100)"))
+		return
+	}
+	tokens, err := model.GetTokensByIdsAndUserId(req.Ids, userId)
+	if err != nil {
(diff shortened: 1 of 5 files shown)


ImogeneOctaviap794:
Closing this PR — the upstream has already merged an equivalent fix in commit 12a48c62 (feat(token): add batch API for fetching token keys). Our changes are fully covered. Thanks!

PR #4228, opened 2026-04-14
Make Claude auto tests mirror Claude Code traffic

nixyme:
## Summary
- make scheduled Claude health checks send Claude Code style headers
- attach `metadata.user_id` to Claude auto-test requests and preserve metadata in Claude conversion
- stop scheduled auto-tests from directly auto-disabling channels while keeping manual test disable behavior intact
- add targeted regression coverage for the Claude Code test fixtures

## Why
A large share of Claude channels were being auto-disabled even though manual Claude Code traffic could still succeed. The root cause was that auto-tests were not close enough to real Claude Code traffic, so Claude-only upstreams were returning false negatives. This change makes the synthetic traffic much closer to real Claude Code requests and removes the most dangerous part of the feedback loop: scheduled tests auto-banning channels.

## Testing
- go test ./controller -run 'TestApplyClaudeCodeTestFixtures'
- go test ./relay/channel/claude -run 'TestRequestOpenAI2ClaudeMessage_PreservesMetadata'
- go test ./controller ./service

## Summary by CodeRabbit

* **Tests**
  * Added test fixtures and coverage for Claude-specific channel testing
  * Added validation for metadata preservation in request transformation

* **Refactor**
  * Enhanced channel test control flow with configurable auto-disable behavior
  * Improved metadata handling in Claude request processing

### DIFF
--- a/controller/channel-test.go
+++ b/controller/channel-test.go
@@ -36,6 +36,21 @@ import (
 	"github.com/gin-gonic/gin"
 )
 
+var claudeCodeTestHeaders = map[string]string{
+	"User-Agent":                                "claude-code/channel-test",
+	"X-Stainless-Arch":                          "arm64",
+	"X-Stainless-Lang":                          "js",
+	"X-Stainless-Os":                            "macos",
+	"X-Stainless-Package-Version":               "channel-test",
+	"X-Stainless-Retry-Count":                   "0",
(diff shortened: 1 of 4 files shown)


seefs001:
newapi不会接受伪造请求特征操作的PR

PR #4581, opened 2026-05-01
fix(relay): restore Claude file and stream status handling

luoyanglang:
## Background

While establishing the local backend baseline, `go test ./...` exposed two regressions in existing tests after the frontend `dist` directories were built for Go embed.

## Problem

The current relay behavior does not match the existing test expectations in two areas:

- OpenAI `file` content converted for Claude is treated through the generic media path, so PDFs and text files can be misclassified as images, while unsupported files are not skipped.
- `StreamScannerHandler` always replaces `RelayInfo.StreamStatus`, which drops any status data that was initialized by the caller before stream handling starts.

- Issue:
- Discussion:
- Related PR:

## Root Cause

The file-source refactor unified media loading but removed the explicit Claude file-content handling path. As a result, filename-derived MIME information was no longer used to distinguish PDFs, text files, images, and unsupported file payloads.

For stream handling, `StreamScannerHandler` unconditionally assigned a new `StreamStatus` instead of only initializing it when the caller had not already provided one.

## Changes

- Restore explicit Claude handling for OpenAI `file` content.
- Convert PDF files to Claude `document` blocks.
- Convert UTF-8 text files to Claude `text` blocks.
- Preserve image file support and skip unsupported file types.
- Preserve preinitialized `RelayInfo.StreamStatus` and only create a new status when it is nil.

## Impact

Affected:

- OpenAI-compatible requests that include `file` content and are relayed to Claude.
- Stream scanner callers that prepopulate `RelayInfo.StreamStatus`.

Not affected:

- Existing text and image URL conversion paths.
- Existing stream status initialization when no status is provided.
- Database schema, billing, authentication, and frontend behavior.

## Testing

- `go test ./relay/channel/claude`
- `go test ./relay/helper`
- `go test ./...`

## Notes

The change is intentionally narrow and keeps the existing test cases as the regression coverage.

## Summary by CodeRabbit

* **New Features**
  * Added support for file attachments in messages to Claude, including PDFs, images, and text files.

* **Bug Fixes**
  * Fixed stream status handling to preserve existing status information instead of overwriting it.

### DIFF
--- a/relay/channel/claude/relay-claude.go
+++ b/relay/channel/claude/relay-claude.go
@@ -1,11 +1,15 @@
 package claude
 
 import (
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io"
+	"mime"
 	"net/http"
+	"path/filepath"
 	"strings"
+	"unicode/utf8"
 
 	"github.com/QuantumNous/new-api/common"
 	"github.com/QuantumNous/new-api/constant"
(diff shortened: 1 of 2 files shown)


seefs001:
1. 针对文件的处理在default分支里
2. 将其重置为新数据是预期行为，大多数AI可能会想顺便把它修改掉，但是这是错误的操作，代码上已经写了注释。

测试失败，是测试用例问题，

PR #4729, opened 2026-05-10
feat: per-model channel testing with test history and model-level enable/disable

wanghualoong:
## Summary

- Upgrade `testAllChannels` to test each model within a channel individually (instead of testing the channel as a single unit), with a 120s per-model timeout
- Record per-model test results in a new `ChannelTestHistory` table for availability tracking (7/15/30 day stats)
- Distinguish **channel-level errors** (invalid API key, account deactivated, insufficient quota) from **model-level errors** — channel-level errors disable the entire channel, model-level errors only disable the affected model
- Add `DisableChannelModel` / `EnableChannelModel` / `IsChannelLevelError` to `service/channel.go`
- Add `UpdateAbilityModelStatus` / `IsAbilityModelEnabled` to `model/ability.go`
- Rebuild channel cache from abilities table (instead of parsing channel.Models string) so per-model disables take effect immediately
- Auto-prune test history older than 30 days
- Auto-re-enable individual models when they pass testing, and re-enable the whole channel when all models recover

## Motivation

The current `testAllChannels` tests each channel with a single model (usually the first one). If a specific model is down but the channel's API key is valid, the whole channel stays enabled — users hitting the broken model get errors. Conversely, if one model fails, the entire channel gets disabled, taking all healthy models offline.

Per-model testing solves both problems: broken models are individually disabled while healthy models continue serving traffic.

## Test plan

- [ ] Run "Test All Channels" from admin UI and verify each model is tested individually
- [ ] Verify `channel_test_histories` table is populated with per-model results
- [ ] Simulate a model-level error (e.g., unsupported model) and verify only that model's ability is disabled
- [ ] Simulate a channel-level error (e.g., invalid API key) and verify the entire channel is disabled
- [ ] Verify auto-recovery: fix a disabled model and run test again — model should be re-enabled
- [ ] Verify test history pruning (records older than 30 days are cleaned up)

## Release Notes

* **New Features**
  * Enhanced channel testing to evaluate each model individually, providing granular reliability assessment instead of single-shot testing.
  * Models can now be automatically enabled or disabled based on test results with improved error classification.
  * Channel test history is now tracked and retained for performance monitoring over time.

[Review Change Stack](https://app.coderabbit.ai/change-stack/QuantumNous/new-api/pull/4729)

### DIFF
--- a/controller/channel-test.go
+++ b/controller/channel-test.go
@@ -901,40 +901,119 @@ func testAllChannels(notify bool) error {
 				continue
 			}
 			isChannelEnabled := channel.Status == common.ChannelStatusEnabled
-			tik := time.Now()
-			result := testChannel(channel, "", "", shouldUseStreamForAutomaticChannelTest(channel))
-			tok := time.Now()
-			milliseconds := tok.Sub(tik).Milliseconds()
 
-			shouldBanChannel := false
-			newAPIError := result.newAPIError
-			// request error disables the channel
-			if newAPIError != nil {
-				shouldBanChannel = service.ShouldDisableChannel(result.newAPIError)
+			models := channel.GetModels()
+			if len(models) == 0 {
+				continue
 			}
 
-			// 当错误检查通过，才检查响应时间
-			if common.AutomaticDisableChannelEnabled && !shouldBanChannel {
-				if milliseconds > disableThreshold {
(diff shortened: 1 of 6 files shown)


wanghualoong:
Closing in favor of #4380, which includes additional features (image model skip, migrateDBFast, channel recovery logic). The improvements from this PR (godoc-style comments, correct stream parameter) will be incorporated into #4380.

PR #4265, opened 2026-04-15
🚀 feat: launch v1.0 — next-generation frontend built from the ground up

t0ng7u:
# ⚠️ 提交说明 / PR Notice
> [!IMPORTANT]
>
> - 请提供**人工撰写**的简洁摘要，避免直接粘贴未经整理的 AI 输出。

## 📝 变更描述 / Description
Introduce a completely redesigned default frontend powered by React 19,
TypeScript, Rsbuild, Radix UI, and Tailwind CSS — replacing the legacy
Semi Design stack with a modern, accessible, and performance-first
architecture.
Highlights:
- Full rewrite: 823 new files, 142K+ lines of purpose-built UI code
- TanStack Router + React Query + React Table for type-safe routing,
  server state, and high-performance data grids
- Zustand for lightweight, scalable state management
- React Hook Form + Zod for declarative, schema-validated forms
- VChart-powered analytics dashboard with real-time metrics
- Complete i18n coverage across 6 languages (en/zh/fr/ja/ru/vi)
- Mobile-first responsive design with adaptive layouts for every page
- Backend-configurable theme switching (default/classic)
- Passkey/WebAuthn, OAuth, and multi-factor authentication flows
- Channel affinity rules, model pricing editor, and subscription management
This marks the transition from v0.x to v1.x — a milestone release
that sets the foundation for the next era of the platform.

## 🚀 变更类型 / Type of change
- [ ] 🐛 Bug 修复 (Bug fix) - *请关联对应 Issue，避免将设计取舍、理解偏差或预期不一致直接归类为 bug*
- [ x ] ✨ 新功能 (New feature) - *重大特性建议先通过 Issue 沟通*
- [ ] ⚡ 性能优化 / 重构 (Refactor)
- [ ] 📝 文档更新 (Documentation)

## 🔗 关联任务 / Related Issue
- Closes # (如有)

## ✅ 提交前检查项 / Checklist
- [ x ] **人工确认:** 我已亲自整理并撰写此描述，没有直接粘贴未经处理的 AI 输出。
- [ x ] **非重复提交:** 我已搜索现有的 [Issues](https://github.com/QuantumNous/new-api/issues) 与 [PRs](https://github.com/QuantumNous/new-api/pulls)，确认不是重复提交。
- [ x ] **Bug fix 说明:** 若此 PR 标记为 `Bug fix`，我已提交或关联对应 Issue，且不会将设计取舍、预期不一致或理解偏差直接归类为 bug。
[...]

PR #4803, opened 2026-05-12
fix(dockerignore): exclude all node_modules + build outputs from build context

xinnyu:
## Problem

The existing `.dockerignore` rule:

\`\`\`
/web/node_modules
\`\`\`

was correct for the **pre-2026 frontend layout** when the React app lived directly under \`web/\`. After PR #4265 split the frontend into \`web/default/\` and \`web/classic/\`, this rule no longer matches any directory — it silently became a no-op.

## Effect

Every \`docker build\` host that previously ran \`bun install\` for local dev iteration (i.e. anyone touching frontend code before building) carries their local \`web/default/node_modules\` (~1.5GB) and optionally \`web/classic/node_modules\` (~700MB) into the build context.

The Dockerfile has three builder stages (\`builder-classic\`, \`builder\`, \`builder2\`), and each issues a \`COPY . .\` that snapshots the entire context into a fresh BuildKit layer blob. Net result on a Mac mini I measured today:

\`\`\`
before this fix:  ~4GB cache written per build (mostly bundled node_modules)
after this fix:   ~1GB cache written per build (only actual compile outputs)
\`\`\`

The in-image \`bun install\` re-installs node_modules anyway, so the bundled junk is pure waste — wasted disk, wasted network if anyone uses remote BuildKit, wasted COPY time.

## Fix

\`\`\`diff
- /web/node_modules
+ **/node_modules
+ /web/default/dist
+ /web/classic/dist
\`\`\`

- \`**/node_modules\` handles all nested locations regardless of future frontend reorganization (no more "silently breaks if path changes")
- Added \`dist/\` exclusions for the same reason (the Dockerfile rebuilds these in-image; bundling stale local builds is the same kind of waste)

## Verification

After this change, on the same Mac mini, a release-mode \`docker buildx build --platform linux/amd64 --push\` writes ~1GB of new BuildKit cache per build instead of ~4GB. Final image content is unchanged (the in-image \`bun install\` produces the same node_modules either way).

## Summary by CodeRabbit

* **Chores**
  * Optimized Docker build configuration to exclude unnecessary dependencies and build artifacts, improving build performance and efficiency.

[Review Change Stack](https://app.coderabbit.ai/change-stack/QuantumNous/new-api/pull/4803)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,5 +7,13 @@ Makefile
 docs
 .eslintcache
 .gocache
-/web/node_modules
+# Exclude all node_modules. The Dockerfile builder stages run `bun install`
+# themselves; bundling the host's local node_modules into the build context
+# is pure waste — adds ~2GB to every `COPY . .` for nothing.
+# (The legacy `/web/node_modules` line below no longer matches anything —
+# after the web/default + web/classic split, the actual dirs are nested.)
+**/node_modules
+# Build outputs — same reasoning; the Dockerfile rebuilds these in-image.
+/web/default/dist
+/web/classic/dist
 !THIRD-PARTY-LICENSES.md


xinnyu:
Closing as duplicate — the .dockerignore exclusions have already been added in #4787 (already merged). My PR proposed a slightly more general pattern (`**/node_modules`) but the explicit paths in #4787 achieve the same result for the current frontend layout. Thanks!

PR #4808, opened 2026-05-12
fix(web/default/playground): hide image-generation models from chat dropdown

xinnyu:
## Problem

The chat playground (`features/playground/`) hardcodes `/v1/chat/completions` as its only endpoint (`API_ENDPOINTS.CHAT_COMPLETIONS` in `constants.ts`). Image-generation-only models like `gpt-image-2`, `dall-e-3`, `imagen-*`, `flux-*` etc. reject chat/completions request shape upstream — they expect `/v1/images/generations`.

**Current UX**: a user with image models available picks `gpt-image-2` in the model dropdown, types a prompt, hits Send, and gets a confusing 400 from upstream like:

> The 'gpt-image-2' model is not supported when using Codex with a ChatGPT account.

Looks like an infrastructure bug to the user but it's actually a fundamental playground/model mismatch.

## Fix

Filter image-only models out of the chat playground's model dropdown via a name-keyword match. List aligned with the `IMAGE_MODEL_KEYWORDS` introduced in #4701 (image workbench PR) for clean drop-in once that lands. Backend `GET /api/user/models` is unchanged — image models still appear in API key creation, image-aware clients, etc.

This is a temporary frontend-only fix until #4701 (or equivalent) ships a proper image workbench page that routes image models through `/v1/images/generations`. When that lands, this filter can be removed (image models would show in the image workbench instead).

## Verification

- Verified locally on a deployment with `gpt-image-1`, `gpt-image-2`, `nano-banana-2` channels: those models no longer appear in playground dropdown; chat-capable models (`gpt-5.3-codex`, `claude-haiku-4-5`, etc.) still appear normally.
- No backend changes — pure frontend filter.

## Summary by CodeRabbit

* **Bug Fixes**
  * Image-generation-only models no longer appear in the chat model dropdown, preventing invalid model selections in the playground.
  * The playground now automatically validates the selected model and falls back to an available model if the current selection becomes invalid.

[Review Change Stack](https://app.coderabbit.ai/change-stack/QuantumNous/new-api/pull/4808)

### DIFF
--- a/web/default/src/features/playground/index.tsx
+++ b/web/default/src/features/playground/index.tsx
@@ -25,6 +25,33 @@ import { usePlaygroundState, useChatHandler } from './hooks'
 import { createUserMessage, createLoadingAssistantMessage } from './lib'
 import type { Message as MessageType } from './types'
 
+// Image-generation-only models that don't accept chat/completions requests.
+// Filtered out of the chat playground's model dropdown so users aren't presented
+// with options that produce confusing upstream errors when "sent". Aligned with
+// upstream PR #4701's image-workbench detection. Drop this list once #4701 lands.
+const IMAGE_MODEL_KEYWORDS = [
+  'gpt-image',
+  'chatgpt-image',
+  'dall-e',
+  'imagen',
+  'flux',
+  'qwen-image',
+  'z-image',
+  'nano-banana',
+  'gemini-2.5-flash-image',
+  'gemini-3-pro-image',
+  'gemini-3.1-flash-image',
+  'gemini-2.0-flash-exp-image',
+  'midjourney',
+  'stable-diffusion',
+  'wanx',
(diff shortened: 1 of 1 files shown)


Calcium-Ion (changes requested):
Thanks for the focused fix, but I don't think we should merge this approach as-is.

Filtering image-generation models out of the playground by name keywords is too coarse and will become a maintenance trap:

- It relies on model-name heuristics rather than endpoint capability metadata, so false positives/negatives are likely as providers add or rename models.
- It hides user-configured models instead of teaching the playground how to route them correctly.
- The playground is expected to gain image-generation support, so this would add behavior that we will need to undo soon.

Please hold this PR and solve the mismatch through proper endpoint/capability handling, or fold it into the upcoming image-generation playground/workbench work.

PR #4871, opened 2026-05-15
fix(auth): expose register_enabled in /api/status and gate sign-up link on login page

Gnonymous:
## Problem

`/api/status` never returned `register_enabled` or `password_register_enabled`, so the sign-in page could not react when an admin toggled the **Registration Enabled** setting in the admin panel.

The \"Sign up\" link on the login page was only gated on `self_use_mode_enabled` — a separate and unrelated concept (single-user self-hosted mode vs. multi-user deployment). This creates two concrete bugs:

1. **Registration disabled, link still visible**: If `RegisterEnabled = false` but `SelfUseModeEnabled = false`, the \"Sign up\" link is shown. Users reach the sign-up page, fill in the form, and get a backend error — a confusing dead end.
2. **Registration enabled, link invisible**: If `RegisterEnabled = true` but `SelfUseModeEnabled = true` (common after the setup wizard selects self-use mode), the link is permanently hidden regardless of the admin setting.

## Root Cause

`controller/misc.go` → `GetStatus()` builds the `/api/status` payload but omits `register_enabled` and `password_register_enabled`. The frontend type definitions (`features/auth/types.ts`) already declare these fields, but since the backend never sends them, they are always `undefined` at runtime and the sign-in page cannot use them.

## Fix

**`controller/misc.go`** — expose two fields in the status response:

```go
"register_enabled":          common.RegisterEnabled,
"password_register_enabled": common.PasswordRegisterEnabled,
```

**`web/default/src/features/auth/sign-in/index.tsx`** — gate the Sign Up link on both settings:

```tsx
// before
{!status?.self_use_mode_enabled && (

// after
{!status?.self_use_mode_enabled && status?.register_enabled !== false && (
```

`!== false` (rather than `=== true`) keeps the link visible for any client that has not yet received the new field (e.g. cached status), preserving backward compatibility.

## Checklist
- [x] Backend: `register_enabled` and `password_register_enabled` now included in `/api/status`
- [x] Frontend: sign-in page respects `register_enabled` when deciding whether to show the Sign Up link
- [x] No new dependencies, no schema changes
- [x] Backward compatible — clients that do not receive the new fields default to showing the link (same behavior as before this fix)

## Bug Fixes
* Sign-in screen now correctly respects registration configuration settings when displaying the sign-up prompt.

[Review Change Stack](https://app.coderabbit.ai/change-stack/QuantumNous/new-api/pull/4871)

### DIFF
--- a/controller/misc.go
+++ b/controller/misc.go
@@ -87,6 +87,8 @@ func GetStatus(c *gin.Context) {
 		"chats":                         setting.Chats,
 		"demo_site_enabled":             operation_setting.DemoSiteEnabled,
 		"self_use_mode_enabled":         operation_setting.SelfUseModeEnabled,
+		"register_enabled":              common.RegisterEnabled,
+		"password_register_enabled":     common.PasswordRegisterEnabled,
 		"default_use_auto_group":        setting.DefaultUseAutoGroup,
 
 		"usd_exchange_rate": operation_setting.USDExchangeRate,
--- a/web/default/src/features/auth/sign-in/index.tsx
+++ b/web/default/src/features/auth/sign-in/index.tsx
@@ -35,7 +35,7 @@ export function SignIn() {
           <h2 className='text-center text-2xl font-semibold tracking-tight sm:text-left'>
             {t('Sign in')}
           </h2>
-          {!status?.self_use_mode_enabled && (
+          {!status?.self_use_mode_enabled && status?.register_enabled !== false && (
             <p className='text-muted-foreground text-left text-sm sm:text-base'>
               {t("Don't have an account?")}{' '}
               <Link


Calcium-Ion (approved):
Approving. Root cause and impact analysis are accurate:

- `/api/status` previously omitted `register_enabled` / `password_register_enabled`, while the default frontend types already declared them, so they were silently `undefined`.
- Gating the sign-up link on `self_use_mode_enabled` alone created two real dead-ends (registration disabled + link visible, registration enabled + link hidden under self-use).

Backend additions in `controller/misc.go` and the corresponding gate in `web/default/src/features/auth/sign-in/index.tsx` are minimal and consistent.

PR #5345, opened 2026-06-06
fix(claude-stream): always emit closing events when upstream sends finish_reason without usage

leeyang1990:
## Problem

When converting OpenAI streaming responses to Claude format, if the OpenAI-compatible upstream sends a chunk with `finish_reason` but no `usage` field, and never follows up with a separate usage-only chunk, the Claude stream is silently truncated — `message_delta` and `message_stop` events are never emitted.

This violates the [Claude Messages SSE protocol](https://docs.anthropic.com/en/api/messages-streaming), which requires every stream to end with `message_stop`. As a result, Claude clients (e.g. Claude Code) hang indefinitely waiting for the stream to terminate.

## Affected Upstreams

This is the protocol behavior of any OpenAI-compatible upstream that omits `usage` when the client doesn't pass `stream_options.include_usage=true`, which is fully compliant with the OpenAI spec. Confirmed reproducers:

- LiteLLM proxy
- Custom OpenAI-compatible gateways
- Some Azure OpenAI deployments

## Root Cause

In `service/convert.go`, the `doneChunk` branch deferred emitting closing events when `usage` was missing, expecting a follow-up usage-only chunk that never arrives:

```go
if oaiUsage == nil {
    oaiUsage = info.ClaudeConvertInfo.Usage
    // Some upstreams emit finish_reason first, then send a final usage-only chunk.
    // Defer closing until usage is available so the final message_delta carries it.
    return claudeResponses  // ← stream silently truncated when no follow-up chunk
}
```

## Fix

1. **`service/convert.go`**: emit `message_delta` + `message_stop` immediately in the `doneChunk` branch, regardless of whether `usage` is available. The `message_delta` event includes usage when available, omits it otherwise — both forms are valid per Claude protocol spec.

2. **`service/convert.go`**: export `BuildClaudeUsageFromOpenAIUsage` and `StopReasonOpenAI2Claude` (as wrappers) so the fallback path in `relay/channel/openai/helper.go` can reuse the conversion logic without duplication.

3. **`relay/channel/openai/helper.go`**: add fallback closing events in `HandleFinalResponse` for cases where the stream is truncated without `finish_reason` at all (e.g. connection dropped, network timeout). Ensures `message_stop` is always emitted.

## Why new-api Should Fix This

The Anthropic Messages SSE spec requires every stream to terminate with `message_stop`:

> The end of the stream is indicated by a `message_stop` event.

new-api is a protocol converter (OpenAI ↔ Claude), and its contract is to accept any *valid* OpenAI input and produce *valid* Claude output. The OpenAI input here is fully spec-compliant — the fix belongs in new-api.

## Testing

Verified locally with:
- LiteLLM proxy → Claude Code: previously hung, now closes correctly
- Direct OpenAI API → Claude Code: still works (no regression)
- Tool calls / streaming with thinking blocks: still works (no regression)

## Release Notes

* **Bug Fixes**
  * Improved reliability of Claude API integration by ensuring proper “stop/close” events are emitted during stream finalization, including fallback closing when needed.
  * Enhanced streaming conversion between OpenAI and Claude to immediately terminate messages when finish reasons arrive, with conditional attachment of usage data.
  * Improved stop-reason mapping to ensure a sensible default is used when an unmapped finish reason is encountered.

### DIFF
--- a/relay/channel/openai/helper.go
+++ b/relay/channel/openai/helper.go
@@ -169,6 +169,36 @@ func HandleFinalResponse(c *gin.Context, info *relaycommon.RelayInfo, lastStream
 		for _, resp := range claudeResponses {
 			_ = helper.ClaudeData(c, *resp)
 		}
+
+		// Fallback: ensure closing events are emitted even if upstream never sent
+		// finish_reason / usage. Some OpenAI-compatible upstreams (e.g. LiteLLM)
+		// truncate the stream without proper terminators, and Claude Code hangs
+		// indefinitely without message_stop.
+		if !info.ClaudeConvertInfo.Done {
+			stopReason := service.StopReasonOpenAI2Claude(info.FinishReason)
+			if stopReason == "" {
+				stopReason = "end_turn"
+			}
+			if usage != nil {
+				_ = helper.ClaudeData(c, dto.ClaudeResponse{
+					Type:  "message_delta",
+					Usage: service.BuildClaudeUsageFromOpenAIUsage(usage),
+					Delta: &dto.ClaudeMediaMessage{
+						StopReason: common.GetPointer[string](stopReason),
+					},
+				})
+			} else {
+				_ = helper.ClaudeData(c, dto.ClaudeResponse{
+					Type: "message_delta",
+					Delta: &dto.ClaudeMediaMessage{
+						StopReason: common.GetPointer[string](stopReason),
+					},
+				})
+			}
+			_ = helper.ClaudeData(c, dto.ClaudeResponse{
+				Type: "message_stop",
+			})
+		}
 		info.ClaudeConvertInfo.Done = true
 
 	case types.RelayFormatGemini:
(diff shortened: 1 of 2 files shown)


leeyang1990:
Thanks @coderabbitai for the review. Pushed [b0707ce04](https://github.com/QuantumNous/new-api/pull/5345/commits/b0707ce04) addressing both points:

1. **Fast-path now also emits terminal events without usage** — `service/convert.go` `SendResponseCount==1` branch was missing the same fallback. It now emits `message_delta` (with or without usage) + `message_stop` consistently with the multi-chunk path.

2. **Fallback closes open content blocks first** — extracted the block-tracking logic into a new exported helper `service.GenerateClaudeStopBlocksForOpenInfo` to avoid duplication. The fallback in `relay/channel/openai/helper.go` now emits `content_block_stop` for any open block before terminal events, keeping the SSE sequence valid.

Compiles cleanly. Verified the same reproducer (LiteLLM → Claude Code with single-chunk finish_reason and no usage) now closes correctly.

kuangzhenfeng:
LiteLLM + new api + claude code 会无法使用，目前完成合并了吗

PR #5384, opened 2026-06-08
fix(stats): include Anthropic cache tokens in usage rankings token_used

broven:
## Problem

The usage rankings / `quota_data` export records `token_used` as `prompt_tokens + completion_tokens` (`model/log.go`, in `RecordConsumeLog`).

- **OpenAI-style usage**: `prompt_tokens` already includes cache-read tokens, so `prompt_tokens + completion_tokens` is the full token volume. ✅
- **Anthropic-style usage**: cache-read (`cache_tokens`) and cache-creation (`cache_creation_tokens`) are reported *separately* and are **not** part of `prompt_tokens`. They were therefore dropped from `token_used` entirely. ❌

For cache-heavy Claude traffic, cache tokens dominate, so the model is massively undercounted in the leaderboard even though its `quota` (billing) is computed correctly from the cache tokens + ratios.

### Observed (production, 24h, one Claude Opus model)

| metric | value |
|---|---|
| rankings `token_used` (prompt+completion) | ~1.96M |
| real cache-read tokens | ~11.7M |
| real cache-creation tokens | ~51.0M |
| real total volume | **~145M** (98%+ cache) |

The leaderboard showed ~1.4% of the real token volume. OpenAI models on the same instance matched their downstream stats exactly, which isolated the issue to the Anthropic-only separate cache accounting.

## Fix

Compute `token_used` in a small provider-aware helper `statsTokenUsage`:

- add cache-read + cache-creation tokens back **only** for `anthropic` usage semantics (where they are excluded from `prompt_tokens`);
- OpenAI-style usage is unchanged (no double counting).

The discriminator is `other["usage_semantic"] == "anthropic"`, which is set in exactly the same place the cache tokens are excluded from `prompt_tokens`, so the two stay in sync. The cache counts are read from the existing `other` payload, so no call-site or struct-signature changes are needed.

This only affects the statistics token count; `quota` / billing is untouched. Historical `quota_data` rows are not rewritten by this change.

## Tests

Added `model/log_stats_test.go` covering OpenAI (no double-count), Anthropic with/without cache tokens (including the real opus-4-8 shape), `float64` payloads, and nil `other`.

```
go test ./model/ -run TestStatsTokenUsage -v
```

## Release Notes

* **Bug Fixes**
  * Improved token usage accounting to correctly calculate and track tokens across different AI providers, with proper support for cache tokens in Anthropic-based integrations.

* **Tests**
  * Added test coverage for token usage calculation scenarios across multiple provider types and edge cases.

### DIFF
--- a/model/log.go
+++ b/model/log.go
@@ -266,9 +266,44 @@ func RecordConsumeLog(c *gin.Context, userId int, params RecordConsumeLogParams)
 		logger.LogError(c, "failed to record log: "+err.Error())
 	}
 	if common.DataExportEnabled {
+		tokenUsed := statsTokenUsage(params)
 		gopool.Go(func() {
-			LogQuotaData(userId, username, params.ModelName, params.Quota, common.GetTimestamp(), params.PromptTokens+params.CompletionTokens)
+			LogQuotaData(userId, username, params.ModelName, params.Quota, common.GetTimestamp(), tokenUsed)
 		})
+	}
+}
+
+// statsTokenUsage returns the total token volume to record for usage statistics
+// (the rankings / quota_data export). It must reflect the real number of tokens
+// processed, regardless of how billing weights them.
+//
+// For OpenAI-style usage, prompt_tokens already includes cache-read tokens, so
+// prompt_tokens + completion_tokens is the full volume. For Anthropic-style usage,
+// cache-read and cache-creation tokens are reported separately and are NOT part of
+// prompt_tokens; they must be added back in, otherwise cache-heavy models (e.g.
+// Claude) are massively undercounted in the rankings even though their quota is
+// computed correctly.
+func statsTokenUsage(params RecordConsumeLogParams) int {
+	tokens := params.PromptTokens + params.CompletionTokens
+	if semantic, _ := params.Other["usage_semantic"].(string); semantic == "anthropic" {
+		tokens += otherTokenCount(params.Other, "cache_tokens")
+		tokens += otherTokenCount(params.Other, "cache_creation_tokens")
+	}
+	return tokens
+}
+
+// otherTokenCount reads a token count stored in the log "other" payload, tolerating
+// the int / int64 / float64 forms it can take across in-memory and decoded maps.
+func otherTokenCount(other map[string]interface{}, key string) int {
+	switch v := other[key].(type) {
+	case int:
+		return v
+	case int64:
+		return int(v)
+	case float64:
+		return int(v)
+	default:
+		return 0
 	}
 }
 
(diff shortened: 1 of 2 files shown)


noir017:
Independent confirmation of this from another deployment, plus one correction worth folding in before merge.

**Same root cause, same magnitude.** On new-api `v1.0.0-rc.24` (revision `5c3abff`), a Claude-native channel over 7 days:

| metric | value |
|---|---|
| `quota_data.token_used` (prompt+completion) | 13,874,185 |
| real cache-read tokens | 2,584,841,713 |
| real cache-creation tokens | 59,209,621 |
| real total volume | **2,657,251,579** |

The dashboard and rankings showed **0.52%** of real volume — close to the 1.4% reported here. OpenAI-semantic models on the same instance (deepseek et al.) matched exactly, isolating it to the Anthropic separate-cache accounting. Billing/`quota` was correct throughout.

**Correction — `cache_write_tokens` is an alias, not an extra component.** Over 10,914 production rows on one model, three keys carried byte-identical sums:

```
cache_creation_tokens     59,209,621
cache_write_tokens        59,209,621
cache_creation_tokens_5m  59,209,621
```

`text_quota.go` writes `cache_write_tokens` as the *normalized cache-creation total* (its own comment says "normalized cache creation total for UI display"; it's the 5m+1h sum, falling back to `cache_creation_tokens`). So summing only `cache_tokens + cache_creation_tokens` as this PR does is correct — but it's worth an explicit test, since `cache_write_tokens` looks like an independent figure and adding it would silently double the cache-creation half. Suggested case:

```go
{
    name: "anthropic cache_write_tokens alias must not double count",
    params: RecordConsumeLogParams{
        PromptTokens: 2, CompletionTokens: 535,
        Other: map[string]interface{}{
            "usage_semantic":           "anthropic",
            "cache_tokens":             28097,
            "cache_creation_tokens":    281100,
            "cache_write_tokens":       281100,  // alias of cache_creation_tokens
            "cache_creation_tokens_5m": 281100,  // ditto
        },
    },
    want: 2 + 535 + 28097 + 281100,
},
```

Also worth noting `input_tokens_total` is not a shortcut here: `text_quota.go` only writes it when the final request format is **not** `RelayFormatClaude`, so it is absent (0) for exactly the Claude-native traffic that needs fixing.

**Rebase note.** This no longer applies cleanly: `LogQuotaData` has since moved to a `QuotaDataLogParams` struct, so the changed line is now `TokenUsed: params.PromptTokens + params.CompletionTokens` inside a struct literal. The `statsTokenUsage`/`otherTokenCount` helpers are unaffected.

Adapted this onto rc.24 and it has been running in production since 2026-08-28 — new `quota_data` rows now track real volume, and `quota` is byte-identical to before. Would be glad to see this land, or a `token stats include cache` toggle if the current accounting is deliberate.

PR #5594, opened 2026-06-18
fix(subscription): apply gateway price to plan payments

lyzgeorge:
## Summary
- Apply the existing Epay top-up payment calculation to Epay subscription checkout
- Apply the existing Waffo Pancake top-up payment calculation to Waffo Pancake subscription checkout
- Store the calculated checkout amount in subscription orders and send it to the gateway

## Why
Subscription plan checkout previously sent `plan.PriceAmount` directly to Epay and Waffo Pancake. This skipped the configured gateway unit price / local-currency ratio used by normal wallet top-ups.

## Notes
- Stripe subscription checkout is unchanged because it uses Stripe Price IDs.
- Creem subscription checkout is unchanged because it is product-price based.
- `plan.PriceAmount` is cast to `int64` to reuse the existing top-up helpers, matching the current integer-oriented recharge amount behavior.

## Testing
- Not run (per request).

## Summary by CodeRabbit

* **New Features**
  * Subscription payment amounts now computed based on user group, enabling differentiated pricing.
  * Added validation to ensure subscription payments meet minimum amount threshold.

### DIFF
--- a/controller/subscription_payment_epay.go
+++ b/controller/subscription_payment_epay.go
@@ -84,10 +84,20 @@ func SubscriptionRequestEpay(c *gin.Context) {
 		return
 	}
 
+	group, err := model.GetUserGroup(userId, true)
+	if err != nil {
+		c.JSON(http.StatusOK, gin.H{"message": "error", "data": "获取用户分组失败"})
+		return
+	}
+	payMoney := getPayMoney(int64(plan.PriceAmount), group)
+	if payMoney < 0.01 {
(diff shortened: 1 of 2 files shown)


lyzgeorge:
I intentionally kept the existing int64(plan.PriceAmount) conversion to reuse the current top-up pricing helpers with the smallest possible change; subscription prices are already treated as whole-number purchase amounts in this flow, and changing the helper signatures to float64 would broaden the scope of the PR beyond this targeted gateway-ratio fix.

PR #5654, opened 2026-06-22
[codex] log image relay upstream channel

neta-zjj:
## What changed

Add a targeted error log in the image relay path when upstream returns a non-200 response.

## Why

The existing error chain only surfaced the upstream body and hid which channel/image path failed. This log now includes:
- `channel_name`
- full `RelayInfo`
- upstream status code
- truncated error text

## Impact

This does not change request/response behavior. It only makes production failures traceable.

## Validation

- `git diff --check`
- `go test ./relay/... ./service/... ./controller/...` could not run here because the local Go toolchain is `go1.18.1`, while `go.mod` requires `1.25.1`.

## Summary by CodeRabbit

* **Chores**
  * Enhanced error logging for relay failures to improve system diagnostics and troubleshooting capabilities.


neta-zjj:
Closed because this change was opened against the wrong forge. The active repository for this work is the TalesofAI Gitea remote.

PR #5895, opened 2026-07-04
fix: use int64 for aggregated usage stats on 32-bit

Nuctori:
## Summary
- widen `quota_data` aggregate counters (`count`, `quota`, `token_used`) to `int64`
- scan flow quota and log statistic aggregates into `int64` targets
- add regression coverage for aggregate totals above 32-bit `int` range

## Why
This is a follow-up to the 32-bit quota overflow fix for ARMv7 environments. User/token quota fields are covered separately, but the dashboard/statistics paths can still exceed `MaxInt32` because they accumulate persisted usage data and SQL `SUM(...)` results.

The change is intentionally scoped to aggregate/stat scan targets and avoids widening single-request billing fields in this PR.

## Tests
- `go test ./model -run "Test(QuotaDataAggregatesSupportInt64Totals|FlowQuotaDataAggregatesSupportInt64Totals|LogStatAggregatesSupportInt64Totals|GetFlowQuotaDataUsesQuotaDataRoleSpecificDimensions|LogQuotaDataSplitsRowsByUseGroupTokenChannelAndNode)$" -count=1`
- `go test ./service -run "TestObserveChannelAffinityUsageCacheByRelayFormat_(MixedMode|UnsupportedModeKeepsEmpty)$" -count=1`

## Notes
`go test ./service -count=1` currently fails in `channel_affinity_usage_cache_test` when the full service package runs, while the named failing tests pass in isolation. This appears to be an existing shared-state/order issue unrelated to this aggregate-int64 change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved quota and usage aggregation to correctly handle very large totals by using wider integer values for counts, quotas, and token usage metrics, ensuring accurate reporting across standard and filtered views.

* **Tests**
  * Added new test coverage for int64-based quota/usage aggregation across standard and flow/group-specific scenarios.
  * Updated existing assertions to match int64 numeric outputs and added a SQL validation test for the identity-only lookup behavior.

### DIFF
--- a/model/aggregate_int64_test.go
+++ b/model/aggregate_int64_test.go
@@ -1,0 +1,115 @@
+package model
+
+import (
+	"testing"
+	"time"
+
+	"github.com/QuantumNous/new-api/common"
+	"github.com/stretchr/testify/require"
+)
+
+func TestQuotaDataAggregatesSupportInt64Totals(t *testing.T) {
+	truncateTables(t)
+
+	require.NoError(t, DB.Create(&QuotaData{
+		UserID:    1,
+		Username:  "alice",
+		ModelName: "gpt-a",
+		CreatedAt: 1000,
+		Count:     1400000000,
+		Quota:     1500000000,
+		TokenUsed: 1200000000,
+	}).Error)
+	require.NoError(t, DB.Create(&QuotaData{
+		UserID:    2,
+		Username:  "bob",
+		ModelName: "gpt-a",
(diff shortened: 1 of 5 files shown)


Calcium-Ion:
Thank you for your pull request. However, we are unable to support 32-bit systems, and we do not have the resources to maintain code for them. Therefore, this PR cannot be merged.

PR #6004, opened 2026-07-08
feat: optional bounded concurrency for batch channel testing

opastorello:
# ⚠️ 提交说明 / PR Notice

> This PR is **AI-assisted**: the code was written with AI help, then reviewed, built and tested by me. The description below is written by me based on my understanding of the change.

## 📝 变更描述 / Description

Batch channel testing (`performChannelTests` in `controller/channel-test.go`) tested every channel **strictly sequentially** — one `testChannel` call at a time. On instances with many channels a run takes very long because each slow/dead upstream is waited on one after another.

This PR makes the batch test run channels through a **bounded worker pool** whose size is configurable, while keeping the default behavior identical to today.

**Backend**
- New setting `MonitorSetting.ChannelTestConcurrency` (`json:"test_concurrency"`), **default `1`** → existing deployments are unchanged unless they opt in.
- Optional env override `CHANNEL_TEST_CONCURRENCY` (mirrors `CHANNEL_TEST_FREQUENCY`/`CHANNEL_TEST_ENABLED`); non-positive values normalize to `1`.
- `performChannelTests` dispatches to at most `N` concurrent goroutines: shared `channelTestSummary` is mutex-guarded, progress is reported through a single serialized `advanceProgress` (monotonic), `ctx` cancellation is honored while acquiring a slot, and `RequestInterval` throttles strictly between dispatches (no trailing wait, no leaked timer).

**Frontend (default theme)**
- Adds a **"Test concurrency"** number field to *Model settings → Routing Reliability*, wired to `monitor_setting.test_concurrency` (mirrors the existing test-interval field), with en/zh strings and `i18n:sync` applied to the other locales.

**Why it works**

Per-channel `testChannel` calls each build their own gin context and recorder, so they are independent. The only shared state (summary counters, progress) is synchronized. Per-channel side effects (`processChannelError`, `EnableChannel`, `UpdateResponseTime`, consume-log writes) operate on independent rows through GORM, which is safe for concurrent use.

## 🚀 变更类型 / Type of change

- [ ] 🐛 Bug 修复 (Bug fix)
- [x] ✨ 新功能 (New feature) — opt-in, backward compatible (default `1`)
- [x] ⚡ 性能优化 / 重构 (Refactor)

## 🔗 关联任务 / Related Issue

- N/A

## ✅ 提交前检查项 / Checklist

- [x] **人工确认:** 描述由我本人撰写整理（AI 辅助编码，已在顶部声明）。
- [x] **非重复提交:** 已搜索现有 Issues/PRs（#5844 仅新增测试超时设置，与本 PR 不重叠）。
- [x] **变更理解:** 我理解这些更改的工作原理及可能影响。
- [x] **范围聚焦:** 本 PR 仅涉及批量测试并发（后端 + 前端设置项）。
- [x] **本地验证:** 后端 `go build`/`go vet`/包测试通过；前端 `bun run typecheck` 通过，改动文件 `oxlint` 0 error。
- [x] **安全合规:** 无敏感凭据；默认值 `1` 保持原有行为。

## 📸 运行证明 / Proof of Work

```
# backend
$ go build ./controller/... ./setting/operation_setting/...   -> ok
$ go vet   ./controller/... ./setting/operation_setting/...   -> ok
$ go test  ./setting/operation_setting/ -run ChannelTestConcurrency -v
--- PASS: TestGetMonitorSetting_ChannelTestConcurrencyNormalizedToAtLeastOne
--- PASS: TestGetMonitorSetting_ChannelTestConcurrencyEnvOverride
--- PASS: TestGetMonitorSetting_ChannelTestConcurrencyEnvNonPositiveNormalizedToOne
$ go test  ./controller/ -run 'SelectChannelsForAutomaticTest|SettleTestQuota'  -> ok

# frontend (web/default)
$ bun run typecheck                       -> tsgo -b, no errors
$ bunx oxlint <changed files>             -> 0 warnings, 0 errors
```

### DIFF
--- a/controller/channel-test.go
+++ b/controller/channel-test.go
@@ -1001,6 +1001,7 @@ func performChannelTests(ctx context.Context, channels []*model.Channel, testUse
 		advanceProgress()
 	}
 
+	dispatched := false
 dispatch:
 	for _, channel := range channels {
 		if ctx != nil && ctx.Err() != nil {
@@ -1009,6 +1010,23 @@ dispatch:
 		if channel.Status == common.ChannelStatusManuallyDisabled {
 			advanceProgress()
 			continue
+		}
+
+		// RequestInterval throttles how fast new tests are dispatched, spacing out
+		// upstream load even when tests run concurrently. Only throttle between
+		// dispatches, never before the first or after the last one.
+		if dispatched && common.RequestInterval > 0 {
+			if ctx == nil {
+				time.Sleep(common.RequestInterval)
+			} else {
+				timer := time.NewTimer(common.RequestInterval)
+				select {
+				case <-ctx.Done():
+					timer.Stop()
+					break dispatch
+				case <-timer.C:
+				}
+			}
 		}
 
 		// Acquire a worker slot, honoring cancellation so a runner that loses its
@@ -1029,20 +1047,7 @@ dispatch:
 			defer func() { <-sem }()
 			testOne(ch)
 		}(channel)
-
-		// RequestInterval throttles how fast new tests are dispatched, spacing out
-		// upstream load even when tests run concurrently.
-		if common.RequestInterval > 0 {
-			if ctx == nil {
-				time.Sleep(common.RequestInterval)
-			} else {
-				select {
-				case <-ctx.Done():
-					break dispatch
-				case <-time.After(common.RequestInterval):
-				}
-			}
-		}
+		dispatched = true
 	}
 
 	wg.Wait()
--- a/controller/channel-test.go
+++ b/controller/channel-test.go
@@ -13,7 +13,6 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/QuantumNous/new-api/common"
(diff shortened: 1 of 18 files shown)


opastorello:
Thanks for the review 🙏 Addressed all three in the latest commit:

1. **Serialize progress callbacks (Major):** all progress reports now go through a single `advanceProgress` helper guarded by a dedicated mutex with a monotonic `completed` counter, so concurrent `testOne` goroutines can no longer report out of order or regress the persisted progress. Replaced the atomic counter accordingly.
2. **Clamp non-positive env override:** `CHANNEL_TEST_CONCURRENCY=0`/`-1` is now applied and then normalized to `1` by the existing clamp, instead of being silently ignored.
3. **Isolate the normalization test:** it now sets `CHANNEL_TEST_CONCURRENCY=""` so ambient CI env can't affect it, plus a new case covering the `env=0 -> 1` normalization.

`go build` / `go vet` / package tests all pass locally.

opastorello:
Applied the nitpick as well: the `RequestInterval` throttle now runs strictly between dispatches (gated on a `dispatched` flag), so there's no trailing wait after the last channel, and it uses a stoppable `time.NewTimer` instead of `time.After` so no timer lingers when `ctx` is cancelled mid-throttle. Thanks for confirming the semaphore/wait-group coordination 👍

PR #6040, opened 2026-07-09
Limit onboarding and invite bonuses by eligibility events

think-back:
## Summary
- Limit new-user signup bonus to the first two accounts per registration IP, while preserving legacy behavior when no registration IP is available.
- Enforce configured email-domain restrictions for registration, email verification, and OAuth user creation.
- Move invite rewards to successful top-up completion and grant a fixed 5 USD to both inviter and invitee.

## Test Plan
- go test -count=1 ./model/...
- go test -count=1 ./controller -run TestRegisterRejectsEmailOutsideDomainWhitelist|TestFindOrCreateOAuthUserRejectsEmailOutsideDomainWhitelist|TestEnsureInitialTokenCreatedDoesNotTriggerInviteReward|TestAddTokenDoesNotTriggerInviteReward|TestAddTokenWithPaymentComplianceUnconfirmedStillDoesNotGrantReward
- git diff --check

## Known unrelated failure
- go test -count=1 ./controller/... still fails at TestStripeCheckoutSessionEmbeddedModeUsesReturnURL because the existing Stripe embedded return URL assertion expects no query params. controller/topup_stripe.go and controller/topup_stripe_test.go are unchanged in this branch.

## Deployment recommendation
- Deploy API/console services that run signup, OAuth, token, quota, and top-up code.
- Router nodes should also be deployed if they share this model/quota/billing code in the production runtime image; otherwise router-only traffic code is not directly changed.


think-back:
Closing because this branch was intentionally based on SolveaCX/new-api:main. The correct PR is https://github.com/SolveaCX/new-api/pull/356.

PR #6070, opened 2026-07-10
fix(stream): preserve usage chunk before non-standard upstream frames

Ankairis:
## Problem

When streaming with upstreams that send non-standard chunks after the usage-bearing `finish_reason` chunk (e.g. OpenCode.ai sends `x-opencode-type` and extra `cost` frames), `OaiStreamHandler` overwrites `lastStreamData` on every frame. `handleLastResponse` then sees a chunk without usage, `containStreamUsage` stays `false`, and the code falls back to `ResponseText2Usage` (local token counting). This causes `cache_tokens` to always be 0 regardless of actual upstream cache hits.

## Root Cause

`relay/channel/openai/relay-openai.go` — `OaiStreamHandler`:

```go
lastStreamData = data   // ← overwritten by non-standard chunk
processTokenData(data)
// ...
handleLastResponse(lastStreamData)  // ← wrong chunk, no usage
```

Upstream stream order:
```
... finish_reason chunk (with usage)   ← has cache_tokens
... x-opencode-type chunk (no usage)   ← overwrites lastStreamData!
... [DONE]
```

## Fix

1. Introduce `usageStreamData` — stores the last chunk that contained valid `Usage`
2. In the stream callback, parse each chunk and save it to `usageStreamData` if it has valid usage
3. Before falling back to `ResponseText2Usage`, try `usageStreamData` first
4. Pass `usageStreamData` to `applyUsagePostProcessing` so the original upstream body is still used for cache-token extraction

## Testing

Verified with upstream curl test that OpenCode.ai returns `prompt_cache_hit_tokens` and `prompt_tokens_details.cached_tokens` correctly in the `finish_reason` chunk. After this fix, those values are preserved instead of being lost to the subsequent non-standard frame.

Closes #(issue if applicable)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved usage reporting for streamed responses by recovering usage data from earlier valid stream updates.
  * Added fallback estimation when usage details are unavailable.
  * Preserved streamed content and trailing response frames while ensuring usage and cached token counts remain accurate in non-standard streaming scenarios.

### DIFF
--- a/relay/channel/openai/relay-openai.go
+++ b/relay/channel/openai/relay-openai.go
@@ -119,6 +119,7 @@ func OaiStreamHandler(c *gin.Context, info *relaycommon.RelayInfo, resp *http.Re
 	var usage = &dto.Usage{}
 	var lastStreamData string
 	var secondLastStreamData string // 存储倒数第二个stream data，用于音频模型
+	var usageStreamData string      // 存储含有 usage 的最后一个 chunk，备用
 	seenStreamToolCalls := make(map[string]struct{})
 	var streamFunctionCallNames []string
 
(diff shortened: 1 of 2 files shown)


huanggefan:
This issue exists in the **opencode** interface.
When stream=True
Finally, an **x-opencode-type** chunk will be returned.

This results in the cache hits recorded by new-api always being 0

Ankairis:
是的，根因就是 OpenCode.ai 在 `finish_reason`（含 usage）之后发了 `x-opencode-type` 非标准块，`lastStreamData` 被覆盖导致 usage 丢失。

这个 PR 的改法是在每个 chunk 处理时独立保存最后一个含有效 `usage` 的 chunk，回退计费前优先用它，同时对非 usage chunk 增加 `strings.Contains("\"usage\"")` 预检避免多余反序列化。

目前已在我这边生产环境验证通过，`cache_tokens` 从恒为 0 正确恢复为实际值。

Ankairis:
Rebased onto the latest `main` — the conflict in `relay/channel/openai/relay-openai.go` is resolved and the change is gofmt-clean.

I also added a regression test (`relay/channel/openai/relay_openai_stream_usage_test.go`) that replays the exact failure shape: a usage-bearing `finish_reason` chunk followed by a non-standard `x-opencode-type` frame. It asserts the real usage and `cached_tokens` are recovered, and I verified it fails on the previous code and passes with this fix.

This has been running in production on my side since the original submission — OpenCode streams now record real `cache_tokens` instead of 0. @seefs001 I'd appreciate a review when you have a moment.

PR #6297, opened 2026-07-18
feat(security): trusted proxies, CORS allowlist, HTTPS security headers

xvyimu:
## Summary
Small, independently reviewable security middleware PR (extracted from the withdrawn large hardening effort).

- **Trusted proxies**: `TRUSTED_PROXY_CIDRS` configures Gin `SetTrustedProxies` (default empty = do not trust `X-Forwarded-For` / `X-Real-IP`).
- **CORS**: replace `AllowAllOrigins` with explicit `CORS_ALLOWED_ORIGINS` allowlist (comma-separated origins; no wildcards). Same-origin deployments need no config.
- **Security headers**: `X-Content-Type-Options`, `Referrer-Policy`, `X-Frame-Options`; `Strict-Transport-Security` only when the request is HTTPS (including `X-Forwarded-Proto` from a reverse proxy).

Default behavior stays safe for single-host / same-origin deploys. Operators behind Cloudflare/Tunnel should set `TRUSTED_PROXY_CIDRS` appropriately (e.g. `127.0.0.1/32` when only local reverse proxy terminates TLS).

## Test plan
- [x] `go test ./middleware ./`
- [ ] Manual: with empty env, no `Access-Control-Allow-Origin: *`
- [ ] Manual: `CORS_ALLOWED_ORIGINS=https://example.com` only allows that origin
- [ ] Manual: HTTPS (or `X-Forwarded-Proto: https`) sets HSTS; plain HTTP does not

## Notes
- No UI / channel / billing changes.
- Does not include session-refresh, SPA routing, adaptive balance, or delivery-seam work (those will be separate PRs).

## Summary by CodeRabbit

* **New Features**
  * Added configurable CORS origin allowlists with credential support and stricter validation.
  * Added standard security response headers, including HTTPS-only HSTS protection.
  * Added trusted reverse-proxy configuration for safer client IP handling.
  * Documented the new security-related environment settings and their defaults.

* **Bug Fixes**
  * Prevented unrestricted cross-origin access when no trusted origins are configured.
  * Rejected invalid or wildcard origin and proxy configurations.

PR #6298, opened 2026-07-18
fix(auth): fail-closed session refresh via user cache

xvyimu:
## Summary
Second small PR from the withdrawn hardening effort: **session auth fail-closed**.

- Session cookies are treated as identity **hints** only.
- On each `UserAuth` / `AdminAuth` / `RootAuth` / `TokenOrUserAuth` session path, reload `username` / `role` / `status` / `group` via **`model.GetUserCache`** (Redis first, DB fallback).
- `UserBase` gains **`Role`** so demotion is enforced without a dedicated `GetUserById` on every dashboard request when cache is warm.
- Ban / missing user → unauthorized/forbidden before the 30-day cookie would expire.
- Unit tests without Redis/DB keep session-only behavior (no panic on `RDB == nil`).

## Test plan
- [x] `go test ./middleware ./model`
- [ ] Manual: disable user in admin → existing browser session loses access on next API call
- [ ] Manual: demote admin → `AdminAuth` routes return insufficient privilege

## Notes
- Does **not** include CORS/HSTS (see #6297), SPA routing, adaptive balance, or delivery seam.
- Quota debit changes from the large branch are intentionally **not** included.

## Summary by CodeRabbit

* **Bug Fixes**
  * Session authentication now normalizes session user IDs and refreshes the user’s identity, role, status, and group before granting access.
  * Sessions for disabled/deleted/missing users are rejected; refresh failures now fail securely when persistence is available.
  * Authorization now validates refreshed status and minimum role requirements rather than relying on stale claims.
  * Cached user snapshots now retain role to keep authorization decisions consistent.
* **Tests**
  * Added coverage for session refresh, fail-closed behavior, admin checks, and context propagation (including status).
* **Documentation**
  * Clarified disabled-vs-other token states for read-only token authentication.

### DIFF
--- a/middleware/auth.go
+++ b/middleware/auth.go
@@ -189,11 +189,18 @@ func authHelper(c *gin.Context, minRole int) {
 				return
 			}
 		} else {
-			username = full.Username
-			// Prefer cached role; fall back to session only if cache is pre-Role schema.
-			if full.Role != 0 {
-				role = full.Role
+			// Authorization always comes from cache/DB snapshot - never fall
+			// back to session role/status. A pre-Role Redis entry leaves Role==0
+			// (zero value); treat that as stale and force a DB reload so we do
+			// not restore a cookie's potentially elevated administrator role.
+			if full.Role == 0 && model.DB != nil {
+				_ = model.InvalidateUserCache(asIntID(id))
+				if dbUser, dbErr := model.GetUserById(asIntID(id), false); dbErr == nil && dbUser != nil {
+					full = dbUser.ToBaseUser()
+				}
 			}
+			username = full.Username
+			role = full.Role
 			status = full.Status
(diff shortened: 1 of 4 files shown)


xvyimu:
﻿Addressed CodeRabbit review on this PR:

1. **Role fail-closed**: never fall back to session/cookie role when cache snapshot has `Role==0`. Invalidate cache and reload from DB; always take `role`/`status` from cache/DB.
2. **Context**: `c.Set("status", status)` is now set on the session auth path (same as `TokenOrUserAuth`).

Pushed: `7f01eeec` on `pr/b2-session-fail-closed`.
Tests: `go test ./middleware` green.

xvyimu:
Withdrawing this contribution. No further action needed from maintainers. Closing and deleting the head branch.

PR #6425, opened 2026-07-23
fix(oauth): stop treating a foreign window.opener as a bind flow

neimaravila:
## 📝 变更描述 / Description

`/oauth/:provider` decides between an account **bind** and a plain **login** with a single signal:

```ts
const mode: 'login' | 'bind' =
  typeof window !== 'undefined' && window.opener ? 'bind' : 'login'
```

`window.opener` cannot carry that meaning. **Any** tab opened from an external link (`target="_blank"`, Slack, Teams, mail clients, another site) has a live opener, and that opener **survives the cross-origin round trip to the identity provider**.

So an ordinary login callback in such a tab is misread as a bind: it posts the bind handshake to a window that speaks no such protocol, renders the "binding your account" screen, and hangs until the 30s deadline in `startOAuthBindResponseDeadline` fires with *"OAuth binding timed out"*. **The backend is never called at all** — `GET /api/oauth/:provider` is never issued.

This is not browser-specific or user-specific: it hits every user who reaches the site through a link instead of typing the URL.

**Root cause reproduced** against a live Keycloak realm:

| step | observed |
|---|---|
| tab opened via `window.open` (= `target="_blank"`) | `window.opener` truthy |
| navigate cross-origin to Keycloak, then back to `/oauth/oidc?code=…` | `window.opener` **still truthy**, `opener.closed === false` |
| resulting mode | `bind` — wrong, this was a login |

### The change

A bind now needs positive proof instead of an inference. The popup opened for a bind is same-origin (`about:blank`) before it is sent to the provider, so it is stamped in its **own** `sessionStorage`. The stamp survives the provider round trip and is scoped to that popup alone, so a login tab can never carry it.

Resolution rules (`resolveOAuthCallbackMode`): a bind requires **both** a live opener **and** our stamp for that exact provider. Everything else resolves to `login`, deliberately — a login callback recovers on its own, whereas a wrongly assumed bind can only time out.

The decision is extracted into `web/src/features/auth/lib/oauth-callback-mode.ts` so it is unit-testable; the storage/opener dependencies are structural, matching the existing style in `oauth-bind-window.ts`.

Affects every provider sharing this callback: OIDC, GitHub, Discord, LinuxDO and custom providers.

## 🚀 变更类型 / Type of change
- [x] 🐛 Bug 修复 (Bug fix)

## 🔗 关联任务 / Related Issue
- No existing issue found; reproduction steps are in Proof of Work below.

## ✅ 提交前检查项 / Checklist
- [x] **人工确认:** Description written and verified by me; AI assistance disclosed below.
- [x] **非重复提交:** Searched open PRs/issues; no existing fix for this callback-mode detection.
- [x] **Bug fix 说明:** Reproduced deterministically; this is a defect, not a design trade-off.
- [x] **变更理解:** Root cause traced end to end, from opener semantics to the 30s handshake deadline.
- [x] **范围聚焦:** 4 files, no unrelated changes.
- [x] **本地验证:** See Proof of Work.
- [x] **安全合规:** No credentials; the stamp holds only a provider slug.

## 📸 运行证明 / Proof of Work

**Reproduction of the root cause** (real Keycloak round trip, browser console):

```
tab opened via window.open      -> openerTruthy: true
after Keycloak round trip       -> openerTruthy: true, openerClosed: false
mode computed by current code   -> "bind"     (should be "login")
```

**Stamp behaves as required** (same browser, same round trip):

```
marker survived round trip      -> "oidc"
marker in the login tab         -> null        (isolated to the popup)
```

**End-to-end against a build of this branch** — the previously hanging scenario (tab with a foreign opener → OIDC callback) now takes the login path immediately:

```
[GET] /api/oauth/oidc?code=…&state=… => [403]
```

`403` is expected here (synthetic `state`); the point is the request is **issued**, where before the fix no request was made and the UI sat on the binding screen for 30s.

**Checks**

```
bun test    -> 86 pass, 0 fail (16 files), including 6 new cases
tsgo -b     -> clean
```

New tests cover: bind popup detected as bind; **login in a tab with a foreign opener stays login** (the regression); marker for another provider ignored; marker without opener; closed opener; storage unavailable.

---

### 🤖 AI assistance disclosure
I am not one of this repository's core contributors. This fix was AI-assisted: the root cause was found by instrumenting the real callback in a browser, and I reviewed the diagnosis, the change and the tests before submitting.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OAuth callback handling by distinguishing login and account-binding flows using the provider and authorization state.
  * Added safeguards for unavailable, blocked, or unwritable browser session storage.
  * Prevented account-binding popups from proceeding when required state tracking fails.
  * Added explicit handling for Telegram account-binding callbacks.

* **Tests**
  * Added coverage for valid, stale, mismatched, missing, and inaccessible OAuth session markers.


Calcium-Ion:
We’re about to squash and merge this PR. Which identity would you prefer for the squashed commit: `Neimar Avila <neimar.avila@vsgroup.com.br>` or your GitHub account, [**[neimaravila](https://github.com/neimaravila)**](https://github.com/neimaravila)?

PR #6531, opened 2026-07-29
feat: SEO improvements, invite rebate fixes, upstream sync

seven7763:
## 📝 变更描述 / Description

本次合并包含以下内容：

### Upstream 同步
- 合并上游 21 个 commit（relaykit 模块拆分、OIDC 自定义名称、CI workflows、per-channel HTTP transport 等）

### SEO 功能完善
- 管理后台 SEO 设置重构为独立 section + RobotsIndex 开关
- sitemap.xml 增加 lastmod/changefreq/priority + 更多路由
- robots.txt 增加更多 Disallow 路径（/token /topup /log /setting 等）
- og:site_name 修复为始终使用品牌短名
- 保存 SEO 设置后立即刷新前端 status 缓存
- controller/seo.go 加 mutex 并发保护 + XML 转义
- 新增 readCachedStatus 共享 helper，去除重复 localStorage 读取

### 邀请返利维护
- 系统任务面板增加 invite_rebate_backfill 类型标签和中文翻译

### 文档站优化
- 配色统一对齐主站（neutral + 蓝色 primary）
- 导航栏与主站完全一致（滚动前透明、滚动后胶囊）
- Hero 组件升级（装饰元素、渐变、更大间距）
- 卡片 hover 效果增强
- 代码块行高优化、表格斑马纹
- 文档助手增加 fenced code block 语法高亮
- DeepChat URL 修正为 deepchat.thinkinai.xyz
- 系统公告 Markdown 渲染修复
- 首页面包屑隐藏

### i18n
- 7 种语言补齐 SEO/任务相关翻译

## 🚀 变更类型 / Type of change
- [x] ✨ 新功能 (New feature)
- [x] 🐛 Bug 修复 (Bug fix)
- [x] ⚡ 性能优化 / 重构 (Refactor)

## ✅ 提交前检查项 / Checklist
- [x] Go build 通过
- [x] Frontend build 通过
- [x] 所有测试通过
- [x] 已部署验证（服务器 152.53.242.23 运行正常）

---
此 PR 由 AI 辅助生成代码。

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added invite top-up rebates with configurable enablement, reward ratios, reporting, leaderboards, logs, and quota transfers.
  * Added user and admin rebate pages, navigation links, and scheduled or on-demand backfill tools.
  * Added configurable SEO settings, dynamic page metadata, social previews, robots.txt, and sitemap.xml support.
* **Bug Fixes**
  * Improved affiliate-code handling by trimming, limiting length, and rejecting invalid or disabled inviters.
  * Prevented disabled accounts and invalid amounts from transferring affiliate quota.
* **Localization**
  * Added translations for invite rebates and SEO settings across supported languages.


seven7763:
## Review follow-up: conflicts resolved, findings addressed

`main` has been merged in (no rebase, so no force-push), the branch is mergeable again, and every CodeRabbit finding is either fixed or answered below. The PR also shrank from **64 files / +6098** to **57 files / +4229** — details at the end.

New commits: `f674bbf` (merge) · `578b4c2` (branding) · `ea93d87` (rebate accounting) · `99eb9b5` (frontend) · `146a066` (docs removal) · `1949a2a` (i18n).

### Conflict resolution

Three files conflicted: `controller/topup.go`, `model/topup.go`, `router/api-router.go`.

The important one is `controller/topup.go`. Since this branch was cut, upstream landed `fix(topup): settle recharge orders atomically` (50e5377) and `fix: harden concurrent quota and status updates` (ccd535e), which introduce `model.RechargeEpay`: row lock, in-transaction pending-status check with `RowsAffected`, `common.QuotaFromDecimalStrict`, and `syncCreditUserQuotaCache` after commit. **I dropped this branch's hand-rolled transaction entirely and took upstream's version**, then moved the rebate hook into `model.RechargeEpay` right after the commit, matching how the Stripe/Creem/Waffo paths already call it.

`model/topup.go` keeps the rebate hook and adopts upstream's `int` quota type; `router/api-router.go` keeps upstream's `UserCriticalRateLimit("aff-transfer")` alongside the new routes.

### Inline comments

| # | Finding | Resolution |
|---|---|---|
| 1 | `model/invite_rebate.go:105` — use conflict-ignore insertion | **Fixed.** Both the skip and grant inserts now go through `insertInviteRebateLedger`, which uses `clause.OnConflict{Columns: topup_id, DoNothing: true}` and branches on `RowsAffected`. `isDuplicateKeyError` is deleted. You were right about PostgreSQL: the violation aborts the transaction, so the credit that follows could never have run. |
| 2 | `model/invite_rebate.go:288-340` — record the amount actually credited | **Fixed.** The grant transaction now re-reads the inviter under `lockForUpdate` and computes the credit with `common.QuotaFromFloatChecked`, per the `AGENTS.md` checked-helper rule. The ledger stores the real credited amount; a fully saturated grant stores `0` with `status=skipped`, the clamp is logged, and `RecordLog` reports the credited value rather than the requested one. Regression test: `TestGrantInviteTopupRebate_LedgerRecordsSaturatedCredit` (partial and full saturation). |
| 3 | `model/invite_rebate.go:309-312` — return a failed ledger cleanup | **Fixed, differently than suggested.** Because the inviter is now re-read and status-checked under a row lock inside the same transaction, the "disabled between read and update" race is gone, so the manual `Delete` is removed rather than error-checked. A non-`1` `RowsAffected` now returns an error, and the rollback removes the ledger row and frees `topup_id` for a retry. Happy to switch to the literal suggested diff if you prefer keeping the delete path. |
| 4 | `model/option.go:333-349` — reset and persist the cutoff | **Fixed.** Every off→on transition now stamps a fresh timestamp, `DB.Save` runs **before** the in-memory flag is published, and its error is returned from `updateOptionMap` instead of discarded. |
| 5 | `web/src/features/invite-rebate/index.tsx:125-138` — `Promise.allSettled` business errors | **Fixed.** `{ success: false }` is handled explicitly. A failed summary now renders an error card with a retry button instead of zero balances and "Rebate disabled". Logs/invitees/affiliate-code failures raise a single aggregated toast — four separate toasts during a full outage seemed worse than one. |
| 6 | `web/src/features/invite-rebate/index.tsx:645-674` — filtering on every keystroke | **Fixed.** Draft inputs (`inviterIdDraft` / `inviteeIdDraft`) are separate from the applied `filters` object that `reload` depends on. Only the Filter button (and Enter) copies draft into applied. |
| 7 | `web/src/lib/seo/apply.ts:154-155` — preserve the real route path | **Fixed.** `path` is passed through unchanged; `robotsIndex` alone controls indexing. Console/auth pages now emit their own canonical/OG URL with `noindex,nofollow`, and the existing post-spread guard still strips the long-tail title off non-home routes. |
| 8 | `controller/topup.go:398-407` — database-idempotent pending→success | **Superseded by the merge.** `model.RechargeEpay` does the conditional update under a row lock and returns `alreadyDone` for duplicate callbacks. |
| 9 | `controller/topup.go:402-405` — checked quota path, cache sync | **Superseded by the merge.** `model.RechargeEpay` uses `QuotaFromDecimalStrict` and calls `syncCreditUserQuotaCache` after commit. |
| 10 | `system-info-section.tsx:119-120` — relative OG image paths | **Fixed.** `SEO.OGImage` now validates through `isOgImageValue` (empty, `/`-prefixed, or a parseable absolute URL) in both schemas — the field's own placeholder is `/logo.png`. Added the error string to all seven locales. |

### Outside-diff comments

| Finding | Resolution |
|---|---|
| `controller/topup.go:391-413` — atomic completion + credit before backfill sees the order | **Superseded by the merge**, same as 8/9: status and quota commit together, and the rebate grant fires post-commit. |
| `system-info-section.tsx:97-99` — preserve the default-enabled robots setting | **Fixed.** Only an explicit `false` / `'false'` normalizes to disabled; a missing option is treated as enabled, matching the browser SEO layer. |

### Minor and nitpick comments

- **Fixed:** invitee-stats and leaderboard-rank queries now propagate errors instead of returning fabricated zero totals or rank 1 (`invite_rebate.go`, both sites) · duplicate `Legal` key removed — this was present in **all seven** locales, not just the three flagged, and was silently overriding the existing translation · `.int()` plus `step={1}` on the rebate ratio and backfill interval · `<TableHead>{t('ID')}</TableHead>` · `line-clamp-1` removed so the rebate details link stays clickable · explicit return types on `useHomeSeo`, the five `dom.ts` helpers, and `RootComponent`.
- **Translated:** all 43 invite-rebate strings in `fr`, `ru`, `ja` and `vi` were shipping their English source text as the value. Now localized, with `{{rate}}` placeholders preserved and verified.
- **Already fixed in d7e496d3:** sitemap `lastmod` removal, and the `systemInfoSchemaWithI18n` URL-validation parity.
- **Docs comments retired:** see below.
- **Not done — `any` casts in `system-info-section.tsx`.** I removed the casts on default-value access (the props type now models the option API's `boolean | string`), but kept `name={'SEO.OGImage' as any}` on the `FormField`s. React Hook Form's `FieldPath` reads a dotted key as a nested path, so a typed field-path contract means either renaming the option keys or adding an adapter — more churn than this PR should carry. Happy to do it in a follow-up if you want it.

### Two things I changed on my own initiative

**Removed a fork's branding that had leaked into upstream defaults.** `common.SystemName`, the default SEO keywords, the `index.html` title / OG / Twitter tags and the document-title fallbacks all carried a downstream product name, and the shell had been switched to `lang="zh-CN"`. All restored to `New API` and English. This conflicted with the protected-project-information rule in `AGENTS.md` and should not have been in the PR — operators who want to rebrand have the configurable SEO layer this PR adds. I also dropped the static `og:url`, since a relative value is not a valid OG URL and the runtime layer sets the real absolute one.

**Removed `docs/superpowers/**` (6 files, ~2000 lines).** These were dated private planning artifacts — design scratch, task checklists, a "ready for next feature" log, local debugging notes. `docs/` upstream is user-facing documentation, so they do not belong here, and they were roughly a third of the diff. This also retires the four review comments about stale `web/default` / `web/classic` paths and the MD058 warning inside them.

### Verification

`go vet ./...` and `go build ./...` clean on both modules · `make test` passes · `bun run typecheck` passes · `bun test` unchanged from `main` on my machine (the same 9 pre-existing failures appear on a clean `main` checkout, seemingly a local Bun version difference — CI pins 1.3.14) · `gofmt` drift this branch had introduced in six files is fixed.

One note: **`bun run typecheck` was failing on this branch before these changes.** `SiteSettings` types `SEO.RobotsIndex` as `boolean | string`, which `SystemInfoSectionProps` rejected. Fixed in `99eb9b5`; the frontend CI job would have failed otherwise.

### Questions for maintainers

1. **Ledger vocabulary for a saturated grant.** When the inviter's `aff_quota` is at the int32 ceiling, I record `status=skipped, rebate_quota=0` so totals stay honest and backfill stops rescanning. Would you rather have a distinct status (e.g. `capped`) to keep `skipped` meaning "not eligible"? It is a one-line change plus a migration note.
2. **Where the rebate hook belongs.** I put `GrantInviteTopupRebate` inside `model.RechargeEpay` after commit, next to `RecordTopupLog`, so all payment providers hook it the same way. If you would rather keep `model/topup.go` free of feature hooks, `RechargeEpay` could return the settled `*TopUp` and the controller could call it instead.
3. **Scope.** This is still two independent features (invite top-up rebate, configurable SEO) in one PR. If that is a barrier, I am happy to split it into two — say the word and I will open them separately rather than have this sit.
4. **CI has not run.** The workflow appears to need maintainer approval for a fork PR; the checks above are from running the same commands locally.

PR #6672, opened 2026-08-05
feat: add Gondola channel

Codonyat:
Adds **Gondola** (https://gondola-ai.com) as a first-class channel type (`61`).

Gondola is an OpenAI-compatible marketplace for Venice AI inference: it routes each request to the cheapest available supplier and exposes one gateway at `https://api.gondola-ai.com`. 260+ models (text, image, video), including an uncensored line. Keys look like `gnd_...`.

### Why a dedicated channel type

Operators can already point a `Custom` channel at Gondola, so this is partly convenience (pre-filled base URL, icon, seed model list). But two of the registrations are functional, not cosmetic:

- **`streamSupportedChannels`** — Gondola supports `stream_options: {"include_usage": true}` and returns a final usage-only chunk. Without the entry, `compatible_handler.go` strips the option and streamed requests fall back to estimated token counts instead of upstream-reported usage. Verified against the live API.
- **`EndpointTypeAnthropic`** — Gondola serves a *native* Anthropic Messages endpoint at `/v1/messages` (streaming and non-streaming, with usage). The adaptor passes Claude-format requests straight through rather than converting them down to `/v1/chat/completions` the way a generic OpenAI channel does. Verified against the live API.

`GET /v1/models` is public and needs no auth, so the default fetch path already works and the channel is added to `MODEL_FETCHABLE_TYPES`.

### Changes

Backend, following the `moonshot` / `New API` (#60) precedent:

- `constant/channel.go` — `ChannelTypeGondola = 61`, base URL, display name.
- `constant/api_type.go` — `APITypeGondola`, inserted before `APITypeDummy`.
- `common/api_type.go` — `ChannelType2APIType` mapping.
- `common/endpoint_type.go` — OpenAI + Anthropic endpoints.
- `relay/common/relay_info.go` — `streamSupportedChannels` entry.
- `relay/channel/gondola/{adaptor.go,constants.go}` — new adaptor. Chat/image delegate to `openai.Adaptor`, Claude format to `claude.Adaptor`; rerank/embedding/audio/gemini/responses report unsupported (Gondola serves none of them: `/v1/completions` and `/v1/rerank` 404, `/v1/embeddings` 501).
- `relay/relay_adaptor.go` — switch branch.

Frontend:

- `constants.ts` — type label, display order, `MODEL_FETCHABLE_TYPES`, key-format hint.
- `channel-type-config.ts` / `channel-utils.ts` — config entry + icon mapping.
- `assets/custom/icon-gondola.tsx` + `lib/lobe-icon.tsx` — custom icon (`@lobehub/icons` has no Gondola glyph), following the `Sub2API` pattern.

`ModelList` is a small seed set; the catalog is a live marketplace, so operators are expected to use "Fetch Models". No `model_ratio.go` defaults, since marketplace prices are dynamic and static entries would be misleading.

### Verification

Ran the CI gate set locally (`GOWORK=off`, with the `web/dist` placeholder):

- `go vet ./...` and `go build ./...` pass for both the root module and `relaykit`.
- `make test` passes except for two failures in `./service/` (`TestObserveChannelAffinityUsageCacheByRelayFormat_MixedMode` and `..._UnsupportedModeKeepsEmpty`) that reproduce identically on an unmodified `main`, so they are unrelated to this change.
- `bun run typecheck` passes. `bun test` shows the same 12 failures on this branch as on unmodified `main` (they are `describe()`-inside-`test()` cases that Bun does not implement on my platform); the 4 channel test files pass, including the 2 new Gondola cases.

Placement of `61` in `CHANNEL_TYPE_DISPLAY_ORDER` is next to OpenRouter (same product category); happy to move it if you would rather it sit elsewhere.

## Summary by CodeRabbit

* **New Features**
  * Added Gondola as a supported channel, including its default base URL, API key guidance, and model discovery.
  * Enabled Gondola for streaming-capable options.
  * Added Gondola request routing for both OpenAI-compatible and Claude Messages styles, plus updated endpoint prioritization.
  * Included Gondola branding via a new UI icon across the interface.
* **Tests**
  * Added Gondola channel tests covering configuration, ordering, icon rendering, base URL validation, and model availability.


Codonyat:
Thanks for the review. Pushed 3186976f for three of the four:

- Split the channel test into one describable behavior per case.
- Explicit `ChannelFormValues` return type on the form fixture.
- `aria-hidden='true'` (plus `focusable='false'`) on the icon, which is decorative next to the channel name.

Leaving two as-is, and happy to change either if you disagree:

**i18n keys for the display strings.** The header of `constants.ts` states that all label/name values *are* the i18n keys and are resolved with `t(value)` at display time, and every sibling follows that: `CHANNEL_TYPES` uses brand literals (`'Sub2API'`, `'New API'`, `'OpenAI'`), `TYPE_TO_KEY_PROMPT` uses English sentences directly (`'Format: APIKey|SecretKey'`), and `channel-type-config.ts` hints do the same (`'DeepSeek API Key'`). `Sub2API` has no locale entry in any file and renders fine via react-i18next key fallback. Introducing separate semantic keys just for Gondola would make it the only channel on a different scheme.

**`currentColor` for the icon stroke.** This is a brand mark, so it keeps its brand color the same way `icon-sub2api.tsx` hardcodes its brand gradient and the `@lobehub/icons` provider glyphs keep theirs. `#c9a96a` is a muted gold that reads on both light and dark backgrounds. Switching to `currentColor` would make Gondola the only monochrome provider icon in the picker.

PR #6706, opened 2026-08-07
relay: emit streamed tool metadata only once

kylefu8:
## Summary

- emit `id`, `type`, and function `name` only on the first Responses-to-Chat tool-call chunk
- keep continuation chunks keyed by `index` with argument deltas only
- omit an empty `type` field instead of serializing it as `null`
- add regressions for long argument continuations and parallel tool calls

## Why

The Responses-to-Chat streaming converter currently repeats the same call ID and type on every function-argument delta. Although the chunks keep the same tool index, some OpenAI-compatible clients interpret each repeated ID as a new tool call. For a large tool payload, one logical call can therefore become many calls; only the first fragment has a function name and later fragments fail as unnamed tools.

This is distinct from the duplicate-index issue fixed in #6225: the tool index is stable, but call metadata is still repeated on every continuation chunk.

OpenAI-style streaming only needs the call metadata on the initial chunk. Subsequent chunks can safely contain the stable index and argument delta.

## Validation

- `cd relaykit && GOMAXPROCS=2 go test -p 2 ./...`
- production-shaped long tool stream: 857 chunks, one tool index, exactly one ID/type/name, and arguments reconstructed as valid JSON
- parallel-tool regression confirms each new tool still receives its own metadata and index
- end-to-end client validation completed a 200-line streamed tool write without producing unnamed or duplicate tool calls

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streamed tool-call responses so identifiers and types appear only on the initial chunk.
  * Prevented repeated tool metadata in subsequent argument updates.
  * Preserved correct metadata handling for multiple parallel tool calls.

* **Tests**
  * Added coverage for streamed tool calls, including incremental arguments and parallel calls.

### DIFF
--- a/relaykit/dto/openai_response.go
+++ b/relaykit/dto/openai_response.go
@@ -123,7 +123,7 @@ type ToolCallResponse struct {
 	// Index is not nil only in chat completion chunk object
 	Index    *int             `json:"index,omitempty"`
 	ID       string           `json:"id,omitempty"`
-	Type     any              `json:"type"`
+	Type     any              `json:"type,omitempty"`
 	Function FunctionResponse `json:"function"`
 }
 
--- a/relaykit/relayconvert/internal/oai_responses/to_oai_chat_resp_test.go
+++ b/relaykit/relayconvert/internal/oai_responses/to_oai_chat_resp_test.go
@@ -1,6 +1,7 @@
 package oairesponses
 
 import (
+	"encoding/json"
 	"testing"
 
 	"github.com/QuantumNous/new-api/relaykit/dto"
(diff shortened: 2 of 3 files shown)


kylefu8:
Additional real-world context: I encountered this while using GPT-5.6 Luna through NewAPI's ChatCompletions-to-Responses compatibility mode in WorkBuddy.

A `show_widget` call with a large SVG payload was streamed as many argument chunks. Because every continuation chunk repeated the same call ID, WorkBuddy interpreted the fragments as separate tool calls. The first fragment was recognized as `show_widget`, while the remaining fragments had empty names and failed with `Tool  not found in agent cli.` After its retries were exhausted, the conversation remained stuck in the thinking state.

After applying this patch, I repeated the same WorkBuddy test with a large `show_widget` payload. WorkBuddy reconstructed and executed one tool call successfully, returned the final response, and exited the thinking state normally. I also validated a separate 200-line streamed `Write` tool call end to end.

PR #6824, opened 2026-08-13
fix(relay): harden gateway against DoS vectors

linseasea:
Several remote DoS vectors reachable by authenticated (and in some cases anonymous) users:

1. **HEIF/ISOBMFF unbounded recursion (stack overflow crash)** — `relay/channel/ali/image.go` recursively parses HEIF/ISOBMFF without depth bound; a crafted image request crashes the process. Add recursion depth + segment count limits.
2. **Auto-ban abuse (whole-fleet DoS)** — channel auto-ban disables an entire channel on a single upstream 429/error-match. Any registered user can poison one request and take down every model on the channel. Add error-count threshold + window before disabling.
3. **Unbounded non-stream upstream response body (OOM)** — relay handlers `io.ReadAll` upstream bodies with no cap; a misbehaving/malicious upstream can OOM the gateway. Add `MAX_RELAY_RESPONSE_MB` (default 64) via `LimitReader`.
4. **Unbounded page_size (DB/memory DoS)** — list endpoints apply `LIMIT` directly from the user parameter; huge values pull the whole table. Cap page size at 1000.
5. **Realtime WS idle connection leak** — `/v1/realtime` connections have no idle timeout/heartbeat; clients can hold connections indefinitely (resource exhaustion, no billing while idle). Add idle timeout + close.

Each fix is minimal and configurable via env where appropriate. Build verified (`go build ./...`).

## Summary by CodeRabbit

* **New Features**
  * Added configurable limits for non-streaming relay responses, defaulting to 64 MB.
  * Added safeguards for pagination values, including maximum page sizes and page numbers.
  * Improved automatic channel disabling by requiring repeated errors across multiple users.

* **Bug Fixes**
  * Successful requests now restore channel health tracking.
  * Realtime connections prevent indefinite idle sessions while supporting active keep-alive traffic.
  * HEIF processing now safely handles deeply nested image metadata.

### DIFF
--- a/common/init.go
+++ b/common/init.go
@@ -177,6 +177,7 @@ func initConstantEnv() {
 	constant.StreamingTimeout = GetEnvOrDefault("STREAMING_TIMEOUT", 300)
 	constant.DifyDebug = GetEnvOrDefaultBool("DIFY_DEBUG", true)
 	constant.MaxFileDownloadMB = GetEnvOrDefault("MAX_FILE_DOWNLOAD_MB", 64)
+	constant.MaxRelayResponseMB = GetEnvOrDefault("MAX_RELAY_RESPONSE_MB", 64)
 	constant.StreamScannerMaxBufferMB = GetEnvOrDefault("STREAM_SCANNER_MAX_BUFFER_MB", 128)
 	// MaxRequestBodyMB 请求体最大大小（解压后），用于防止超大请求/zip bomb导致内存暴涨
 	constant.MaxRequestBodyMB = GetEnvOrDefault("MAX_REQUEST_BODY_MB", 128)
--- a/common/page_info.go
+++ b/common/page_info.go
@@ -8,8 +8,9 @@ import (
 
 // MaxPageSize caps the page_size accepted from query parameters. List
 // endpoints use PageSize directly as SQL LIMIT; an unbounded value would let a
-// single request pull every matching row into memory (F-37).
-const MaxPageSize = 1000
+// single request pull every matching row into memory (F-37). GetPageQuery
+// clamps every page size to this value.
+const MaxPageSize = 100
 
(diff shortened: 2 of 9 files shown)


linseasea:
Thanks for the review. All actionable comments have been addressed in 91b249e:

- **page_info**: unified the page-size cap — `MaxPageSize` is now the single clamp (100) used by `GetPageQuery`.
- **relay.go**: the channel error streak is now cleared on every successful path, including `RelayTask`.
- **api_request.go**: the response-size cap now fails closed via a `limitReadCloser` (reads past the limit return an error) and `Close` is delegated to the original body.
- **relay_realtime.go**: both client and target WebSocket connections are closed after the terminal select so the other reader unblocks promptly.
- **channel.go**: streak deletion is now atomic with streak identity (`CompareAndDelete`) so a stale holder can no longer delete a newer streak.

linseasea:
Hi @Calcium-Ion @seefs001 — gentle ping on this security-hardening PR stack (#6824 / #6825 / #6826). They have been ready since Aug 13, and every CodeRabbit comment was addressed on Aug 17 (build + relaykit tests verified).

Quick recap of why they matter:

- #6826 masks upstream credentials in every relayed error path — streaming chunks, non-stream errors, task/video/audio, Midjourney relay, /v1/alpha/search and response headers. Without it, misbehaving upstreams can leak channel API keys straight to end clients.
- #6825 closes quota-bypass paths — upstreams omitting `usage` currently settle quota as 0 across many providers, plus image `n` multiplier and the unset-ratio admin gate.
- #6824 hardens against remote DoS — HEIF recursion crash, auto-ban abuse, unbounded relay body OOM, page_size, and realtime idle-connection leak.

Happy to rebase, split, or restructure anything that would make review easier. Thanks for your time!

PR #6825, opened 2026-08-13
fix(billing): close quota-bypass paths (usage fallbacks, image n multiplier, unset-ratio gate)

linseasea:
A set of billing-integrity fixes. When an upstream omits `usage`, several handlers settle quota as 0 → the request is effectively free (billed to the operator). Also closes related pricing gaps:

- **Usage fallbacks** — add token-count fallbacks when upstream omits usage for: rerank (ali/SiliconFlow/Jina/Xinference), embeddings (MokaAI/Ollama), zhipu + cohere, Responses API (non-stream), Claude (non-stream, incl. AWS Bedrock delegation), Dify (non-stream), xunfei (stream + non-stream), AWS Nova, ollama chat (stream + non-stream), AWS native non-stream.
- **Image `n` multiplier** — image-generation requests with `n>1` were billed as a single image in Replicate-style paths; apply the multiplier.
- **`accept_unset_model_ratio_model` admin gate** — ordinary users could self-enable acceptance of unset model ratios and use unpriced models; restrict the toggle to admins.
- **Stream goroutine hygiene** — cohere/zhipu/xunfei streaming used bare goroutines + unbuffered channels; connection drops leaked goroutines (also a slow DoS). Tied to the same files as the zhipu/cohere/xunfei fallbacks.

Build verified (`go build ./...`).

## Summary by CodeRabbit

- **Security**
  - Restricted enabling the Accept Unset Ratio Model setting to administrators and higher roles.

- **Usage Reporting**
  - Improved token usage estimates when providers omit or return zero usage data across chat, embedding, rerank, and compaction requests.
  - Enhanced consistency across streamed and non-streamed responses.

- **Reliability**
  - Improved stream shutdown behavior when clients disconnect.

- **Billing**
  - Image requests now account for the requested image count in applicable pricing calculations.

### DIFF
--- a/controller/user.go
+++ b/controller/user.go
@@ -1498,7 +1498,11 @@ func UpdateUserSetting(c *gin.Context) {
 		NotifyType:                       req.QuotaWarningType,
 		QuotaWarningThreshold:            req.QuotaWarningThreshold,
 		UpstreamModelUpdateNotifyEnabled: upstreamModelUpdateNotifyEnabled,
-		AcceptUnsetRatioModel:            req.AcceptUnsetModelRatioModel,
+		// F-27: only admins may enable accept-unset-ratio. Self-service by a
+		// regular user bypasses the "model price not configured" gate, letting
+		// them use deliberately unpriced models at the default ratio (37.5x),
+		// which is both an availability-control and pricing-control bypass.
+		AcceptUnsetRatioModel: user.Role >= common.RoleAdminUser && req.AcceptUnsetModelRatioModel,
 		RecordIpLog:                      req.RecordIpLog,
 	}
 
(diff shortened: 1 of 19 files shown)


linseasea:
Thanks for the review. All actionable comments have been addressed in 2bd497e:

- **claude/aws**: usage fallback now runs inside `HandleClaudeResponseData` before the response is serialized, so clients see the same usage settlement bills; the duplicated post-write fallback blocks were removed.
- **cohere/xunfei**: stream completion is now delivered via channel close — `[DONE]` is rendered only after all buffered frames are consumed (no more dropped/out-of-order terminal signal).
- **cohere non-stream**: fallback uses `ResponseText2Usage` so completion tokens are estimated, not just prompt.
- **ollama stream**: the usage estimate is applied before the final usage frame is emitted.
- **zhipu**: the terminal signal send is now blocking-with-cancel (no dropped completion); non-stream fallback uses `ResponseText2Usage`.
- **image_handler**: the `n` multiplier is only applied when the adaptor reported zero usage, preventing double billing for adaptors that already embed the image count.

PR #6826, opened 2026-08-13
fix(relay): mask upstream credentials in every relayed error path

linseasea:
Security hardening: when an upstream echoes the channel API key back in an error (common with free/aggregator upstreams and misbehaving proxies), the relay forwarded it **verbatim** to the client. This masks credentials across every relayed error surface:

- **Streaming chunks (F-13/F-20)** — SSE error chunks and `message`-only chunks are parsed and masked before relay; bare key formats (`sk-`, `gsk_`, `hf_`, `xai-`, `AIza`, JWT, `pplx-`/`nvapi-`/`r8_`/GitHub/GitLab tokens, `Bearer <v>`, `invalid key <v>`, service-account fields) are covered, including JSON-escaped quoting.
- **Non-stream errors (F-20)** — `code/type/param` and whole serialized OpenAI/Claude errors are masked, not just `message`.
- **Task/video/audio errors (F-42)** — `TaskErrorWrapper` only masked when the message contained post/dial/http; now masks unconditionally; `TaskErrorFromAPIError` masks too.
- **Midjourney relay (F-64)** — `/mj/submit/*`, `/mj/task/:id/image-seed`, and persisted `Description`/`FailReason` (replayed via task fetch) relayed raw upstream bodies without masking.
- **`/v1/alpha/search` (F-65)** — 200 bodies were `io.Copy`-ed to the client without masking.
- **Response headers (F-66)** — upstream `Authorization`/`X-Api-Key`/`Api-Key` etc. in response headers were copied to the client; now blocked (with F-34's header blocklist).

A regression fix for the F-20 masking (enum field over-masking) is included. Build + `relaykit` tests verified.

## Summary by CodeRabbit

- **Security**
  - Sensitive credentials, tokens, private keys, and authentication details are now masked in errors, streamed responses, task results, and relayed content.
  - Credential-bearing response headers are blocked from being forwarded.
  - Sensitive values are sanitized across supported provider formats, including JWTs and bearer tokens.

- **Bug Fixes**
  - Improved protection against accidental exposure of upstream secrets in realtime streams, API errors, search responses, and image-generation workflows.
  - Upstream response read failures now return a clear error instead of forwarding incomplete content.

### DIFF
--- a/common/str.go
+++ b/common/str.go
@@ -141,3 +141,9 @@ func MaskEmail(email string) string {
 func MaskSensitiveInfo(str string) string {
 	return kitutil.MaskSensitiveInfo(str)
 }
+
+// MaskSensitiveKeys masks only credential-shaped values (no URL/domain/IP
+// rewriting), for enum-like fields where full masking would mangle values.
+func MaskSensitiveKeys(str string) string {
+	return kitutil.MaskSensitiveKeys(str)
+}
--- a/relay/alpha_search_handler.go
+++ b/relay/alpha_search_handler.go
@@ -95,8 +95,18 @@ func AlphaSearchHelper(c *gin.Context, info *relaycommon.RelayInfo) (newAPIError
 	if contentType := httpResp.Header.Get("Content-Type"); contentType != "" {
 		c.Writer.Header().Set("Content-Type", contentType)
 	}
+	// F-65: read + mask the upstream body before relaying. /v1/alpha/search
+	// used to io.Copy the raw body to the client, bypassing the F-13/F-20
+	// masking system; an upstream that echoes the channel Authorization value
+	// in a 200 body would leak the key. (F-36 already caps body size here via
+	// DoApiRequest's LimitReader.)
+	responseBody, err := io.ReadAll(httpResp.Body)
+	if err != nil {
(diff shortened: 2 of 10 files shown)


linseasea:
Thanks for the review. All actionable comments have been addressed in 11ad545:

- **Critical**: `/v1/alpha/search` and Midjourney 2xx/persisted payloads now use `MaskSensitiveKeys` (credential-only) so resource URLs survive; `MaskSensitiveKeys` now also covers service-account fields and PEM blocks.
- **stream_scanner**: the usage-triggered recursive walk no longer traverses assistant content fields (`content`/`delta`/`text`).
- **mask.go**: credential prefixes and key-value field names are anchored with word boundaries (no more `task-…`/"monkey" false positives); JWT/bearer anchored too.
- **error.go**: `ToOpenAIError`/`ToClaudeError` mask fields individually — Message via `MaskSensitiveInfo`, enum-like fields via `MaskSensitiveKeys`.
- **service/error.go**: `TaskErrorFromAPIError` uses the nil-guarded `apiErr.Error()` accessor.
- **tests**: `mask_formats_test.go` now asserts exact expected outputs per case instead of a tail heuristic.

PR #6949, opened 2026-08-20
fix(relay): bound the wait for upstream response headers (fixes unbounded heap growth → OOM)

txgo:
Fixes the memory growth reported in #6947. Very likely also the root cause of #6731,
which reported the same symptom (production OOM on `/v1/responses` after ~64h) but was
closed for template reasons — that report is what made me look for a shared mechanism
rather than treat ours as a local quirk.

## The problem

`newRelayHTTPTransport()` bounds the dial, the TLS handshake and expect-continue — but
**nothing bounds how long it waits for the upstream response headers** after the request
has been written:

```go
transport = &http.Transport{
    Proxy:                 http.ProxyFromEnvironment,
    DialContext:           dialer.DialContext,      // 30s ✅
    ForceAttemptHTTP2:     true,
    TLSHandshakeTimeout:   10 * time.Second,        // 10s ✅
    ExpectContinueTimeout: time.Second,             //     ✅
    //                       ResponseHeaderTimeout  ❌ unset → wait forever
}
```

An upstream that accepts the connection and then never answers — **without sending
FIN/RST**, which is what happens when a NAT/firewall silently drops the flow or the
provider hangs — parks the goroutine in `net/http.(*persistConn).roundTrip` forever.

That goroutine keeps the whole request alive, so **three copies of the request body stay
reachable for the lifetime of the process**: the raw bytes from `io.ReadAll` in
`CreateBodyStorageFromReader`, the decoded messages held as `json.RawMessage`, and the
re-marshalled upstream body from `common.Marshal`.

`BodyStorageCleanup` cannot help here — it is registered correctly, but it runs *after*
`c.Next()` returns, and for these requests `c.Next()` never returns.

## Evidence (v1.0.0-rc.23, production, ~1000 users)

**23 goroutines** stuck in `persistConn.roundTrip` on a single 40h-old instance, blocked
between **353 and 1894 minutes** (5.9h to 31.5h), all entered through the relay:

```
goroutine 63245 [select, 1894 minutes]:
net/http.(*persistConn).roundTrip(...)      /usr/local/go/src/net/http/transport.go:2911
net/http.(*Client).Do(...)
github.com/QuantumNous/new-api/relay/channel.doRequest(...)
github.com/QuantumNous/new-api/relay/channel.DoApiRequest(...)
github.com/QuantumNous/new-api/relay.ResponsesHelper(...)
github.com/QuantumNous/new-api/controller.Relay(...)
```

**96.9% of the live heap**, sampled *after a forced GC* (`/debug/pprof/heap?gc=1`), sits in
those three body copies — 892 MiB surviving three GC cycles, `HeapObjects` dropping 30×
while bytes dropped only 25% (i.e. what survives is all large buffers).

The floor tracks uptime — same image, same config, same load:

| uptime | live heap after forced GC |
|---:|---:|
| 0.1 h | 33.7 MiB |
| 13.8 h | 89.2 MiB |
| 40.1 h | 510.0 MiB |
| 146.8 h | 955.2 MiB |
| **172.9 h** | **OOMKilled (137)** |

Doubling the memory limit and adding `GOMEMLIMIT` only moved the OOM from 132h to 172.9h
(+31%). Full write-up, `pprof -traces` retention chain and reproduction steps are in #6947.

## The fix

One effective line, plus a configurable knob:

```go
if common.RelayResponseHeaderTimeout > 0 {
    transport.ResponseHeaderTimeout = time.Duration(common.RelayResponseHeaderTimeout) * time.Second
}
```

Three things worth flagging for review:

**1. Why not `RELAY_TIMEOUT`.** That sets `http.Client.Timeout`, which covers the whole
response read and would cut legitimate long streaming calls — which is exactly why it
defaults to `0`. `ResponseHeaderTimeout` only bounds the wait for the headers; **streaming
after the headers arrive is unaffected**.

**2. Why the default is 1800s and not something tight.** Non-streaming upstreams usually
send the response headers only once generation has finished, so this value has to leave
room for a long completion. 1800s is **12× shorter than the shortest hang observed here**
while leaving several times the headroom a normal non-streaming request needs. `0` restores
the previous unbounded behaviour. Happy to change the default if you'd prefer something
more conservative — the mechanism matters more than the number, since any finite value
turns "held forever" into "held for at most N minutes".

**3. Why the assignment is outside the `else` branch.** `newRelayHTTPTransport()` normally
takes the `http.DefaultTransport.Clone()` path, so the literal in the `else` block is rarely
executed — and `http.DefaultTransport` does not set `ResponseHeaderTimeout` either. Putting
it next to the other `transport.*` lines covers both paths.

This repo already sets `ResponseHeaderTimeout` on its other outbound transports
(`controller/model_sync.go:100`, `controller/ratio_sync.go:200`), so this looks like the
relay path was simply missed rather than a deliberate choice.

## Changes

| file | what |
|---|---|
| `service/http_client.go` | set `ResponseHeaderTimeout` on the relay transport |
| `common/constants.go` | new `RelayResponseHeaderTimeout` var + doc comment |
| `common/init.go` | `RELAY_RESPONSE_HEADER_TIMEOUT`, default `1800` |
| `service/http_client_response_header_timeout_test.go` | covers both the configured value and `0` = disabled |
| `README.md`, `.env.example` | document the new variable |

`go build ./...`, `go vet`, and `go test ./common/... ./service/...` all pass.

I run this in production and can supply further profiles or long-run data on request.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable timeout for waiting for upstream response headers.
  * Defaults to 1,800 seconds and can be disabled by setting the value to `0`.
  * Streaming remains unaffected once response headers arrive.
  * Large timeout values are safely handled to prevent unexpectedly short timeouts.

* **Documentation**
  * Documented the new configuration option, supported values, and behavior in the environment example and README.


txgo:
感谢 @coderabbitai 的评审。两条都看了,**一条全采纳、一条部分采纳并说明理由**(commit `8d8dfd3e`):

**① `common/init.go:113` 的取值校验 —— 部分采纳。**

**溢出这一半是真问题,已修**:秒数超过约 9.2e9 时 `time.Duration` 会溢出,可能**回绕成一个极小的正值** ——
那会把**所有**转发请求都打断,而不只是卡死的那些。现在转换前先 clamp,并补了负值与溢出两条回归测试。

**「负值/非法值直接启动失败」这一半没做**,理由两条:
- 现有的 `if seconds > 0` 已把负值当作「关闭」处理;
- **本文件邻近的同类项还更宽松** —— `RelayIdleConnTimeout` 是无任何 guard 直接
  `time.Duration(x) * time.Second`。在这里单独引入 fail-startup 会与文件其余部分不一致。

如果维护者希望在全仓统一推行 env 校验,我乐意配合改 —— 但那更像是独立的一次改动,不宜夹在本修复里。

**② 测试改用 testify —— 全采纳。** 核实过 `service/` 下**所有**测试文件都用 testify,我原来的
`t.Fatalf` 确实不合惯例,已改为 `require.Equal` / `require.Zero` / `require.Positive`。

现共 4 条子测试:配置生效 / `0`=关闭 / 负值=关闭 / 溢出被 clamp。
`go build`、`go vet`、`go test ./common/... ./service/...` 均通过。

Calcium-Ion (approved):

PR #7015, opened 2026-08-25
Fix Playground quick-start media model handoff

think-back:
## Summary

- Keep image and video quick-start prompts available during a locked model handoff while carrying the explicitly selected media model into generation.
- Prefer Seedance 2.5 for video quick starts, with capability-aware fallback to visible authorized models.
- Preserve website PE model selections such as `seedance-2.5` instead of silently falling back to Veo.

## Verification

- `web/default`: `bun test src/features/playground` — 162 pass, 0 fail
- `web/default`: `bun run build:check` — passed
- `website`: `bun test src/components/cli-prompt-action-panel.test.tsx` — 1 pass, 0 fail
- `website`: `bun run build` — passed
- Targeted ESLint for changed files — passed
- `git diff --check` — passed

Full repository lint still has pre-existing errors outside this change; the changed files pass targeted lint.


think-back:
Closing this accidentally targeted upstream PR. The intended review target is SolveaCX/new-api main, whose fork baseline differs from upstream main; a correctly scoped PR will be opened there.

PR #7026, opened 2026-08-26
Fix Grok text OAuth auto-refresh after 401

think-back:
## Summary

- Refresh expired Grok Subscription OAuth credentials after the first replayable text-request 401.
- Persist refreshed credentials with the existing channel refresh lease and key CAS, then rebuild auth/CLI headers and replay once.
- Sanitize refresh failures and second 401 responses; keep image/media writes non-replayable.

## Verification

- go test -p 1 ./relay/channel/groksubscription -count=1
- go test -p 1 ./relay/channel -count=1
- go build ./relay/channel/groksubscription
- git diff --check

## Notes

- No production deployment or service restart was performed.
- Full relay/channel tree has unrelated baseline failures and linker memory pressure; those are outside this diff.


think-back:
Superseded by the clean-base PR in the SolveaCX repository: https://github.com/SolveaCX/new-api/pull/884

PR #7037, opened 2026-08-27
Fix Playground attachment restoration after refresh

think-back:
Playground refreshes replaced browser-local messages with sanitized server snapshots, so uploaded inline image/video attachments disappeared. This merges local attachment data back into matching same-conversation server message versions while keeping base64 out of persistence. Verified with bun test ./src/features/playground, bun run typecheck, and bun run lint.


think-back:
Superseded by clean PR #893 against the active SolveaCX main branch; this branch was based on an obsolete frontend history.

PR #7228, opened 2026-09-06
feat: per-channel TTFB (first-token) timeout with automatic fallback

HClSuzku:
## Summary

Adds a per-channel **TTFB (first-token) timeout** for streaming requests with automatic fallback.

**Problem:** When an upstream channel sends response headers but never delivers the first data chunk, the request would previously stall until the much longer overall streaming timeout — and even then it wouldn't fall back through the relay's channel retry logic in a timely manner. In practice this turns one slow upstream into multi-minute hangs for every user request routed to it.

**Solution:** After upstream response headers arrive, start a timer for the configured TTFB window. If the first data chunk doesn't arrive in time, the request fails immediately with a channel-level error (`channel:ttfb_timeout`, HTTP 502) so the relay falls back to the next available channel. Nothing has been written to the client yet at that point, so the fallback is safe and transparent.

## Changes

- `common/constants.go`, `common/init.go`: new global `TTFB_TIMEOUT` env var (seconds, `0` = disabled)
- `relay/helper/stream_scanner.go`: TTFB timer armed when headers arrive, disarmed on first data chunk; per-channel `ttfb_timeout` overrides the global default
- `relay/channel/openai/relay-openai.go`, `relay/channel/openai/relay_responses.go`: surface TTFB timeout as a channel error to trigger fallback
- `relay/common/stream_status.go`: new `StreamEndReasonTTFBTimeout`
- `relaykit/dto/channel_settings.go`, `relaykit/types/error.go`: new `ttfb_timeout` channel setting + `ErrorCodeChannelTTFBTimeout`
- `web/src`: TTFB timeout input in the channel advanced settings drawer (frontend form schema, defaults, settings JSON builder)

## Notes

- Timing starts when headers arrive, **not** at request start — normal streams are completely unaffected
- `0`/unset disables the feature (preserves existing behavior)
- Tested in production: a channel with a slow first token now fails fast and the relay falls back to the next channel as expected

## Summary by CodeRabbit

- **New Features**
  - Added configurable time-to-first-byte (TTFB) timeouts for streaming requests.
  - Set a global default with `TTFB_TIMEOUT` or override it for individual channels.
  - Added channel settings UI for choosing the timeout in seconds; `0` disables the limit.
  - If the first data arrives too late, the request returns a channel error and can automatically fall back to the next channel.
  - Prevented keepalive pings from being sent before the first upstream data arrives.
  - Improved accessibility by associating numeric setting labels with their input controls.

### DIFF
--- a/common/constants.go
+++ b/common/constants.go
@@ -173,6 +173,14 @@ var RelayIdleConnTimeout int // unit is second
 // ResponseHeaderTimeout only bounds the wait for the response headers; once the
 // headers arrive, streaming is unaffected.
 var RelayResponseHeaderTimeout int // unit is second
+
+// TTFBTimeout is the global default first-token (TTFB) timeout in seconds for
+// streaming requests: once the upstream response headers have arrived, if the
+// first data chunk does not arrive within this window the request fails as a
+// channel error ("channel:ttfb_timeout") so the relay can fall back to the
+// next channel. 0 disables it. Per-channel override:
+// ChannelSettings.TTFBTimeoutSeconds (channel setting JSON field "ttfb_timeout").
+var TTFBTimeout int // unit is second
 var RelayMaxIdleConns int
 var RelayMaxIdleConnsPerHost int
 
--- a/common/init.go
+++ b/common/init.go
@@ -112,6 +112,7 @@ func InitEnv() {
 	RelayTimeout = GetEnvOrDefault("RELAY_TIMEOUT", 0)
 	RelayIdleConnTimeout = GetEnvOrDefault("RELAY_IDLE_CONN_TIMEOUT", 90)
 	RelayResponseHeaderTimeout = GetEnvOrDefault("RELAY_RESPONSE_HEADER_TIMEOUT", 1800)
+	TTFBTimeout = GetEnvOrDefault("TTFB_TIMEOUT", 0)
 	RelayMaxIdleConns = GetEnvOrDefault("RELAY_MAX_IDLE_CONNS", 500)
 	RelayMaxIdleConnsPerHost = GetEnvOrDefault("RELAY_MAX_IDLE_CONNS_PER_HOST", 100)
 
--- a/relay/channel/openai/relay-openai.go
+++ b/relay/channel/openai/relay-openai.go
@@ -142,6 +142,18 @@ func OaiStreamHandler(c *gin.Context, info *relaycommon.RelayInfo, resp *http.Re
 		}
 	})
 
+	// TTFB timeout: the upstream sent response headers but never delivered the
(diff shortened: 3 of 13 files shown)


HClSuzku:
Addressed all 4 review points in 7a0c143.

**1. (Major) Ping during TTFB window breaks fallback** — fixed. `StreamScannerHandler` now tracks `firstChunkArrived` (atomic bool); the ping goroutine drops ticks until the first real upstream data chunk arrives, so no bytes hit the client while the TTFB timer can still trigger channel fallback. Once the first chunk lands, ping resumes normally (write path unchanged). Added `TestStreamScannerHandler_PingSuppressedBeforeFirstChunk`: first chunk delayed 1.5s past a 1s ping interval, asserts the chunk is forwarded and no `: PING` ever reaches the body. Full package tests pass.

**2. (Minor) `extraSettingsConfigured` misses `ttfb_timeout`** — fixed in `channel-mutate-drawer.tsx`: `ttfb_timeout` is now watched and included in the extra-settings-configured check, so the "unconfigured" notice no longer shows when TTFB is the only custom setting.

**3. (Minor) `NumericSpinnerInput` drops FormControl a11y props** — fixed: the component now accepts and forwards `id`, `aria-describedby`, `aria-invalid` onto the focusable element (edit input or display button, mutually exclusive states), so label `htmlFor` and aria wiring from `FormControl` land correctly.

**4. (Minor) TTFB help text "0 disables" is misleading** — fixed: now reads "0 uses the global TTFB_TIMEOUT default".

Verification: `go test ./relay/helper/` green, web `tsgo -b` typecheck + production build pass, amd64 binary compiled and deployed.

HClSuzku:
Round-2 comments addressed in `eae940c`:

1. **errcheck (ignored pipe errors)** — the writer goroutine now captures `pw.Close()` and both `fmt.Fprint` errors, reports them through a channel, and the test fails if any write failed.

2. **Scheduling race in the ping-suppression test** — the 1.5s delay used to start counting from goroutine launch, so a late-scheduled handler could let the first chunk arrive before the ping ticker even existed (false pass). The writer now waits on a `signalReadCloser` that closes `scannerStarted` on the scanner's first `Read`, i.e. only after the handler is genuinely consuming the upstream body. Verified with `go test -race ./relay/helper/` on the targeted tests.

3. **Label not associated with the control** — when a `label` is provided, the local `Label` now sets `htmlFor`, and the control id falls back to a stable `useId()`-generated value when the caller doesn't supply one, so the label always points at the focusable input/button.

Note: full-package `-race` runs still flag a pre-existing data race in `logger.logHelper()` (logger.go:115) caused by parallel tests logging concurrently — unrelated to this PR, and the non-race suite (what CI runs) is fully green.

PR #7463, opened 2026-09-19
feat(models): expose upstream max_model_len on /v1/models

### LINKED ISSUE #5847: 配置和通过v1/models返回模型上下文窗口
## 提交前必读（请勿删除本节）

- 文档：https://docs.newapi.ai/
- 使用问题先看或先问：https://deepwiki.com/QuantumNous/new-api
- 开启透传后的转发相关反馈不接受 issue；透传模式会直接转发请求，请自行确认上游行为。
- 不接受 coding plan、逆向渠道等技术支持类 issue。
- 警告：删除本模板、删除小节标题或随意清空内容的 issue，可能会被直接关闭；重复恶意提交者可能会被 block。

**您当前的 newapi 版本**

当前版本：v1.0.0-rc.14

**提交确认**

[//]: # (方框内删除已有的空格，填 x 号)
- [x] **非重复 issue:** 我已搜索现有 [Issues](https://github.com/QuantumNous/new-api/issues?q=is%3Aissue)，确认目前没有类似 issue。
- [x] **提交前必读:** 我已完整阅读上方“提交前必读”，并已查看文档 https://docs.newapi.ai/、项目 README 且向 AI 提问，确认这不是使用、配置或接入类问题，且现有版本无法满足需求。
- [x] **模板完整:** 我未删除此模板中的任何引导内容或小节标题，并会按要求完整填写。
- [x] **维护成本:** 我理解项目维护者精力有限，不遵循模板要求的 issue 可能会被无视或直接关闭。

**功能描述**
当前版本在/v1/models中只返回了自定义模型名称，未携带上下文窗口数据，导致claudecode等工具按照工具本身默认值调用，多轮对话后出现上下文超出模型设置大小调用出错。

检查上级vllm的v1/models输出，包含模型上下文

{"object":"list","data":[{"id":"qwen3.6","object":"model","created":1782959108,"owned_by":"vllm","root":"Qwen/Qwen3.6-35B-A3B-FP8","parent":null,"max_model_len":225280,"permission":[{"id":"modelperm-854535ba81a5847c","object":"model_permission","created":1782959108,"allow_create_engine":false,"allow_sampling":true,"allow_logprobs":true,"allow_search_indices":false,"allow_view":true,"allow_fine_tuning":false,"organization":"*","group":null,"is_blocking":false}]}]}

检查newapi的v1/models输出，仅有模型名称

{"data":[{"id":"qwen3.6","object":"model","created":1626777600,"owned_by":"openai","supported_endpoint_types":["openai"]}],"object":"list","success":true}

**应用场景**
使用场景

hariag:
## Problem

vLLM and SGLang advertise each model's maximum context window as `max_model_len` in their OpenAI `/v1/models` payload, but the user-facing listing served by new-api drops the field entirely: `buildOpenAIModel` rebuilds the list locally from the pricing configuration and never asks the upstream engine what it serves.

Clients that auto-configure context management from the models listing (e.g. Claude Code) therefore cannot detect the real limit — see #5847. On our deployment the engine reports `max_model_len: 1048576` on `http://<vllm>/v1/models`, while the gateway's user-facing response carries no context field at all.

## Solution

A scheduled system task (`upstream_context_sync`) probes every **enabled vLLM/SGLang channel's** upstream `/v1/models` once per minute, caches the advertised limit per user model name, and `buildOpenAIModel` surfaces it as `max_model_len` on the user-facing `GET /v1/models`.

### Design notes

- **Runs inside the existing scheduled system task framework** (`service/system_task.go`): the DB lease deduplicates execution across master nodes and the runner supplies a cancellable context — no bespoke background goroutine or shutdown wiring.
- **Only engine channel types are probed** (`ChannelTypeVLLM`, `ChannelTypeSGLang`) — the only providers that report `max_model_len`. Other providers are never contacted, so their credentials are not attached to pointless periodic requests.
- **Base URLs are validated** exactly like the channel status probe (`controller/channel_inference.go`): http/https only, non-empty host, no userinfo/query/fragment. Redirects are not followed, so channel keys never leak to other hosts.
- **`model_mapping` is honoured**: a probe result for the engine-side name is mapped back to every user-facing name that can route to it (served names, mapping chains, cycle-safe with an 8-hop bound).
- **Smallest bound wins** when several channels serve the same user model — it is the only limit every serving channel can honour.
- **Cache semantics**: a probe reply counts as healthy even when it advertises no `max_model_len` (an engine that dropped the field clears stale values). If *every* upstream probe fails, the previous cache is kept as last-known-good. Unknown models omit the field (`omitempty`), so responses stay byte-identical to the current behaviour for channels that never advertise it.
- **Bounded resources**: at most 8 concurrent probes, 4 MiB response cap per probe (`limit+1` explicit overflow check), 10 s per-probe timeout, bounded worker pool per pass.

### Configuration

| Env | Default | Meaning |
|---|---|---|
| `UPSTREAM_CONTEXT_SYNC_FREQUENCY` | `60` | Sync interval in seconds; `0` disables the sync task |

## Testing

- New tests in `controller/upstream_model_context_test.go` (16 test functions): collection, SGLang parity, non-engine channels ignored, cross-channel minimum, `model_mapping` alias resolution, unsafe base URL rejection, oversized-response rejection, pre-flight **and** mid-flight cancellation, dead-channel tolerance, last-known-good on total outage, stale-value withdrawal when the engine stops advertising, mapping-chain/cycle/8-hop-bound unit tests, env-driven schedule.
- `go build ./...`, `go vet ./...`, `gofmt` clean; `go test ./controller/ ./model/ -count=1` passes; focused tests pass with `-race`.
- End-to-end smoke test with a real binary: SQLite + a fake vLLM upstream — user token `GET /v1/models` returns `"max_model_len": 1048576`, chat relay and Anthropic-format endpoints show no regression, unauthenticated requests still 401.

Fixes #5847

## Summary by CodeRabbit

* **New Features**
  * Model listings now include the upstream maximum context length when available.
  * Context limits are synchronized automatically for supported vLLM and SGLang channels.
  * Model aliases and mappings are accounted for, using the smallest available limit.

* **Reliability**
  * Synchronization tolerates unavailable channels and preserves the last known values when all upstream checks fail.
  * Secure upstream connections and response validation help prevent unsafe synchronization requests.

* **Tests**
  * Added coverage for context-limit discovery, mappings, validation, scheduling, failures, and cancellation.


hariag:
Both findings addressed in two follow-up commits (pushed to this branch):

- 3ee6109 — **CWE-319**: probes with a non-empty channel key now require `https` base URLs and are skipped otherwise; unauthenticated HTTP probes (empty/"EMPTY" key) remain supported, along with the existing URL restrictions (no userinfo/query/fragment) and redirect blocking.
- 05ea496 — **per-channel last-known snapshots**: successful probe results are stored per channel with observation timestamps; failed probes retain that channel's previous snapshot; snapshots are refreshed on recovery, pruned for channels that become ineligible or answer healthy without the field, and the aggregate cache is rebuilt under the existing RWMutex by taking the smallest limit across current routed names (including `model_mapping` chains).

Verified: `go build ./...`, focused tests, full `go test ./controller/ ./model/ -count=1` regression, `go vet ./...`, focused tests with `-race`, gofmt clean.

